PR #10, opened 2017-11-30
Fix discarding of non-duplicate observations occurring before a duplicate

### LINKED ISSUE #9: Investigate SQLAlchemy's nested transations
There is some evidence that suggests our nested transactions are not behaving properly. The ``moti_infill.py`` script is reporting that it's making insertions, however those insertions never actually appear in the database.

The transaction management in this package is "complicated". We haven't pinned our version of ``SQLAlchemy`` in the ``requirements.txt`` so there is a real possibility that the behaviour of nested transactions has changed out from under us. We need to:

1. Upgrade and pin to the latest stable version of ``SQLAlchemy`` and
1. Ensure that the nested transactions are behaving the way that they should

rod-glover:
Resolves #9 
Also fixes a minor (non-problem producing) bug due to a typo.

### DIFF
--- a/crmprtd/moti.py
+++ b/crmprtd/moti.py
@@ -145,7 +145,7 @@ def check_history(stn_id, sesh):
         with sesh.begin_nested():
             hist = History(station = stn)
             sesh.add(hist)
-    except Excetion as e:
+    except Exception as e:
         log.error('History_id could not be found or created for native_id {}'.format(stn_id), exc_info=True)
         raise e
     log.debug('Created history_id {}'.format(hist.id))
--- a/crmprtd/moti.py
+++ b/crmprtd/moti.py
@@ -98,6 +98,8 @@ def process_observation_series(sesh, os):
 
             o = Obs(time=t, datum=float(value), variable=var, history=hist)
 
+            # Regarding the following non-canonical but working code, see
+            # https://github.com/pacificclimate/crmprtd/issues/9#issuecomment-348042673
             try:
                 with sesh.begin_nested():
                     sesh.add(o)
--- a/crmprtd/moti.py
+++ b/crmprtd/moti.py
@@ -102,6 +102,7 @@ def process_observation_series(sesh, os):
                 with sesh.begin_nested():
                     sesh.add(o)
                 successes += 1
+                sesh.commit()
                 log.debug("Inserted {}".format(o))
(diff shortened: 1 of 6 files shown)


rod-glover:
@jameshiebert Tests fail only in Python 3.3 environment. Package `py==1.5.2` (latest) is incompatible with Python 3.3. Shall we ditch testing in Python 3.3? I'm not sure which version of py is compatible with it, nor whether it would be a good idea to pin to it. I favour ditching. You?

rod-glover:
@jameshiebert Also would value your thoughts on the SQLAlchemy [weirdness](https://github.com/pacificclimate/crmprtd/issues/9#issuecomment-348042673) I documented. I could develop a minimal demo case of this if desired.

rod-glover:
Regarding replacing `pytest-capturelog` with `pytest-catchlog`, I ran into [this issue](https://github.com/eisensheng/pytest-catchlog/issues/36) (more or less), and adopted the simplest applicable solution. See most recent comments to it.

jameshiebert:
From [here](https://www.python.org/downloads/release/python-337/):

> Python 3.3.7
> 
> Release Date: 2017-09-19
> 
> Python 3.3.x has reached end-of-life. This is its final release. It is a security-fix source-only release.

Probably safe to drop support for it.

crmprtd/moti.py
```diff
@@ -143,7 +147,7 @@ def check_history(stn_id, sesh):
         with sesh.begin_nested():
             hist = History(station = stn)
             sesh.add(hist)
-    except Excetion as e:
+    except Exception as e:
```
jameshiebert: Oh jeez. Good catch.

tests/conftest.py
```diff
@@ -205,49 +207,115 @@ def moti_sawr7110_xml():
...
     <product operational-mode="official">
-      <title>Observation from BC Meteorological Stations
-      </title>
-      <field>meteorological
-      </field>
-      <category>observation
-      </category>
-      <creation-date refresh-frequency="PT1H">2013-11-29T13:22:57-08:00
-      </creation-date>
+      <title>Observation from BC Meteorological Stations</title>
```
jameshiebert: Thanks for the formatting. I'm guessing that we just copy/pasted this from a download originally, but this is far easier to read.

PR #13, opened 2018-06-01
Resolves #11, added wmb tests and cov to travis

### LINKED ISSUE #11: Add coverage tests to continuous integration testing
While test coverage isn't a perfect metric, it *can* be a good indicator as to how completely the code base is tested. The ``crmprtd`` test coverage is *very poor*:

~~~
----------- coverage: platform linux, python 3.5.2-final-0 -----------
Name                        Stmts   Miss  Cover
-----------------------------------------------
crmprtd/__init__.py            20     17    15%
crmprtd/compat.py               4      1    75%
crmprtd/db.py                  44      3    93%
crmprtd/ec.py                 180     30    83%
crmprtd/moti.py               149     37    75%
crmprtd/wamr.py               165     48    71%
crmprtd/wmb.py                235    235     0%
crmprtd/wmb_exceptions.py      15     15     0%
-----------------------------------------------
TOTAL                         812    386    52%
~~~

For this issue, please enable coverage testing in our [TravisCI continuous integration config](https://github.com/pacificclimate/crmprtd/blob/master/.travis.yml). Configure `coverage` to only measure coverage on modules in the ``crmprtd`` package directory and configure it to fail our test suite under a generous threshold. Let's start with 50% (so that our preset tests pass) and then let's crank that up to 80-90 as we slowly add tests to cover the rest of the code base.

nikola-rados:
Changed script command in **.travis.yml** to test for coverage and added tests for **wmb.py**. The script command will fail if the tests cover below 50% of the source code.  With the addition of the new test file **test_wmb.py** the total coverage for crmprtd is 70%.

### DIFF
--- a/tests/test_wmb.py
+++ b/tests/test_wmb.py
@@ -1,151 +1,93 @@
-import sys
 import pytest
 import pytz
 import csv
-import random
 from sqlalchemy import and_
 
 from datetime import datetime
-from decimal import Decimal
 from io import StringIO
-from tempfile import TemporaryFile
 
 from pycds import Obs, History, Network, Station
 from crmprtd.wmb import ObsProcessor, insert_obs, check_history
 
 
 @pytest.mark.parametrize(('val', 'hid', 'd', 'vars_id', 'expected'), [
-    # It looks like the vars table is indexed?
-    (2.7, 20, datetime(2016, 9, 13, 6, tzinfo=pytz.utc), Decimal(1), 4)
+    (2.7, 20, datetime(2016, 9, 13, 6, tzinfo=pytz.utc), 1, 4)
 ])
 def test_insert_obs(test_session, val, hid, d, vars_id, expected):
     insert_obs(val, hid, d, vars_id, test_session)
+
     q = test_session.query(Obs)
     assert q.count() == expected
-    # sanity check
-    q = test_session.query(Obs).filter(Obs.datum == val, Obs.history_id == hid, Obs.time == d, Obs.vars_id == vars_id)
-    assert q.count() == 1
+
+    q = test_session.query(Obs.datum).filter(Obs.history_id == hid, Obs.time == d, Obs.vars_id == vars_id)
+    result = next(iter(q[0]))
+    assert result == val
 
 
-@pytest.mark.parametrize(('pre_insert', 'post_insert', 'row_check'), [
-    (6, 9, 1)
-])
-def test_check_and_insert_stations(test_session, pre_insert, post_insert, row_check):
-    lines = '''station_code,weather_date,precipitation,temperature,relative_humidity,wind_speed,wind_direction,ffmc,isi,fwi,rn_1_pluvio1,snow_depth,snow_depth_quality,precip_pluvio1_status,precip_pluvio1_total,rn_1_pluvio2,precip_pluvio2_status,precip_pluvio2_total,rn_1_RIT,precip_RIT_Status,precip_RIT_total,precip_rgt,solar_radiation_LICOR,solar_radiation_CM3
-11,2018052711,.00,14.2,55,10.4,167,81.160995,2.1806495,5.5260615,.00,.00,,,.00,.00,,.00,.00,.00,.00,,.0,
-11,2018052712,.00,16.4,57,9.1,152,81.667679,2.166688,5.4912086,.00,.00,,,.00,.00,,.00,.00,.00,.00,,.0,
-11,2018052713,.00,16.9,54,11.3,185,82.228363,2.5902824,6.5181026,.00,.00,,,.00,.00,,.00,.00,.00,.00,,.0,
-11,2018052714,.00,17.8,53,10.5,185,82.773972,2.6630962,6.9062028,.00,.00,,,.00,.00,,.00,.00,.00,.00,,.0,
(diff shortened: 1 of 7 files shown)


tests/test_wmb.py
```diff
@@ -0,0 +1,151 @@
+import sys
```
jameshiebert: While it's nice that you started digging into the test coverage, that wasn't necessarily part of this task. We can keep it for now, but for the future, try to be wary of "scope creep".

tests/test_wmb.py
```diff
@@ -0,0 +1,151 @@
...
+from io import StringIO
+from tempfile import TemporaryFile
+
+from pycds import Obs, History, Network, Station
+from crmprtd.wmb import ObsProcessor, insert_obs, check_history
+
+
+@pytest.mark.parametrize(('val', 'hid', 'd', 'vars_id', 'expected'), [
+    # It looks like the vars table is indexed?
+    (2.7, 20, datetime(2016, 9, 13, 6, tzinfo=pytz.utc), Decimal(1), 4)
```
jameshiebert: Why are you using `Decimal` type here?

tests/test_wmb.py
```diff
@@ -0,0 +1,151 @@
...
+
+    o = ObsProcessor(test_session, data, 1000)
+
+    o.network_id = test_session.query(Network.id).filter(Network.name == 'FLNRO-WMB')
+    o.network = test_session.query(Network).filter(Network.id == o.network_id).first()
+
+    stations = set()
+    id = 0
+    q = test_session.query(Station.native_id).join(Network).filter(Network.id == o.network_id).first()
+    for record in q:
```
jameshiebert: Not sure what you're doing here. Your query only selects one record (by using `.first()`), but then you iterate over the (one) record? That's... strange. And why are you converting the native id to an integer and then back to a string? [native_ids are strings](https://github.com/pacificclimate/pycds/blob/master/pycds/__init__.py#L82). Even if they look like a number, they are strings.

tests/test_wmb.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+@pytest.mark.parametrize(('val', 'hid', 'd', 'vars_id', 'expected'), [
+    # It looks like the vars table is indexed?
+    (2.7, 20, datetime(2016, 9, 13, 6, tzinfo=pytz.utc), Decimal(1), 4)
+])
+def test_insert_obs(test_session, val, hid, d, vars_id, expected):
+    insert_obs(val, hid, d, vars_id, test_session)
+    q = test_session.query(Obs)
+    assert q.count() == expected
```
jameshiebert: So I'm not sure that there's any value to parametrizing the "expected" value. Isn't the result 4, just the number of observations that are already in the test_session when you get it (3) plus the number that you have inserted (1)? In that case, that number will never change with any other parameters to this test. What may be more valuable is to count the observations in the database before and after running `insert_obs()` and assert that the count after == the count before + 1.

tests/test_wmb.py
```diff
@@ -0,0 +1,151 @@
...
+    # sanity check
+    q = test_session.query(Obs).filter(Obs.datum == val, Obs.history_id == hid, Obs.time == d, Obs.vars_id == vars_id)
```
jameshiebert: This is a fine sanity check. The other way to do it would be to filter on history_id, time, and vars_id and then assert that datum == val (since datum is essentially the one item that is recorded... i.e. the "independent variable").

tests/test_wmb.py
```diff
@@ -0,0 +1,151 @@
...
+    data = []
+    f = maybe_fake_file(lines)
+    reader = csv.DictReader(f)
+    for row in reader:
+        data.append(row)
+
+    o = ObsProcessor(test_session, data, 1000)
+
+    o.network_id = test_session.query(Network.id).filter(Network.name == 'FLNRO-WMB')
+    o.network = test_session.query(Network).filter(Network.id == o.network_id).first()
```
jameshiebert: You actually don't need the `network_id` here at all. You can just say:

```python
o.network = test_session.query(Network).filter(Network.name == 'FLNRO-WMB')
```

The ORM will do the work for you of using the `id` column to track the relationships.

tests/test_wmb.py
```diff
@@ -0,0 +1,151 @@
...
+    assert q.count() == pre_insert
+
+    o.check_and_insert_stations(stations)
+
+    q = test_session.query(Station)
+    assert q.count() == post_insert
+
+    test_val = random.sample(stations, 1)
+    q = test_session.query(Station).join(Network).filter(and_(Station.native_id == test_val[0], Network.id == o.network_id))
+    assert q.count() == row_check
```
jameshiebert: Again, why parametrize row_check? Can't you just say 1 here?

tests/test_wmb.py
```diff
@@ -0,0 +1,151 @@
...
+
+    q = test_session.query(Station)
+    assert q.count() == pre_insert
+
+    o.check_and_insert_stations(stations)
+
+    q = test_session.query(Station)
+    assert q.count() == post_insert
+
+    test_val = random.sample(stations, 1)
```
jameshiebert: Do you have so many stations that you can't simply test all of them? Testing should be fast; just do them all :) Using randomness in your tests actually makes them non-deterministic and making them more challenging to re-run.

jameshiebert:
The follow-up commits clean this up a lot. It looks great! I'll go ahead and merge.

PR #16, opened 2018-06-11
Resolves #12, code refactor

### LINKED ISSUE #12: Reorganize modules
I've started a [dev branch](https://github.com/pacificclimate/crmprtd/tree/dev) for reorganizing the code into a four stage pipeline. See the description in [`crmprtd/__init__.py`](https://github.com/pacificclimate/crmprtd/blob/dev/crmprtd/__init__.py) for a description of the stages.

For this issue, begin with a checkout of the `dev` branch, and begin to pull almost all code out of the executable scripts except for command line argument handling and delegation to a external functions. Basically this lets us ignore any testing coverage in `scripts/` because there should be no business logic there.

Move download specific code into the `[network_name]/download.py` modules and try to get a handle on how to separate the downloading code from the normalization code (in some networks, they are a bit intertwined at the moment).

nikola-rados:
Code from the scrips/ directory was moved into the allotted crmprtd/[network name] where it was further split into download.py and normalize.py. The download portion of the pipeline has been isolated however, the rest has been placed into normalize for the time being.

### DIFF
--- a/crmprtd/wmb_dir/normalize.py
+++ b/crmprtd/wmb_dir/normalize.py
@@ -1,0 +1,53 @@
+# Standard module
+import sys
+import logging
+
+# debug
+
+# Local
+from crmprtd.wmb import ObsProcessor, DataLogger
+
+# Installed libraries
+from sqlalchemy import create_engine
+from sqlalchemy.orm import sessionmaker
+
+def wmb_normalize(args, log, data):
+    dl = DataLogger()
+
+    # Open database connection
+    # I think this would fall under the align/insert?
+    try:
+        engine = create_engine(args.connection_string)
+        Session = sessionmaker(engine)
+        sesh = Session()
+        sesh.begin_nested()
+    except Exception as e:
+        dl.add_row(data, 'db-connection error')
+        log.critical('''Error with Database connection
+                            See logfile at {l}
+                            Data saved at {d}
+                            '''.format(l = args.log, d = data_archive), exc_info=True)
+        sys.exit(1)
+
+    try:
+        op = ObsProcessor(sesh, data, args)
+        op.process()
+        if args.diag:
+            log.info('Diagnostic mode, rolling back all transactions')
+            sesh.rollback()
+        else:
+            log.info('Commiting the session')
+            sesh.commit()
+
+    except Exception as e:
+        dl.add_row(data, 'preproc error')
+        sesh.rollback()
+        data_archive = dl.archive(args.archive_dir)
+        log.critical('''Error data preprocessing.
+                            See logfile at {l}
+                            Data saved at {d}
+                            '''.format(l = args.log, d = data_archive), exc_info=True)
+        sys.exit(1)
+    finally:
+        sesh.commit()
+        sesh.close()
--- a/crmprtd/wmb_dir/normalize.py
+++ b/crmprtd/wmb_dir/normalize.py
@@ -23,16 +23,16 @@ def save_file(reader, cache_dir, data):
         copier.writerows(data)
 
 
-def reader2data(reader, cache_dir):
+def reader2data(reader):
     data = []
     for row in reader:
         data.append(row)
-    save_file(reader, cache_dir, data)
     return data
 
 
 def prepare(args, log, reader):
-    data = reader2data(reader, args.cache_dir)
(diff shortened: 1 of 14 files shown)


scripts/hourly_wmb.py
```diff
@@ -6,174 +6,16 @@
...
-        # It's non-ideal but neither classes support coroutine send/yield
-        lines = []
-        def callback(line):
-            lines.append(line)
-
-        self.connection.retrlines('RETR {}'.format(self.filename), callback)
-
-        r = csv.DictReader(lines)
-        return r
+    run(args)
```
jameshiebert: At this point, you've refactored everything down so well that ``main()`` doesn't actually do anything. That's good... it means we can get rid of it! I'd recommend either changing the import to ``from crmprtd.wmb_dir.download import run as main``, or directly calling ``run()`` instead of ``main()``. I'd prefer the latter.

And you can apply this suggestion to each of the other scripts as well.

scripts/real_time_ec.py
```diff
@@ -19,99 +19,11 @@
 # Local
 from crmprtd import retry
 from crmprtd.ec import makeurl, ObsProcessor, parse_xml, extract_fname_from_url
+# needs import for ec_download
```
jameshiebert: This doesn't really need to be said. If we import something, we assume that it's needed :)

scripts/wamr_hourly.py
```diff
@@ -87,46 +81,7 @@ def main():
...
-                datetime.now(), '%Y-%m-%dT%H-%M-%S'))
-        with open(args.cache_file, 'w') as cache_file:
-            cache_rows(cache_file, rows, fieldnames)
-
-    log.info('{0} observations read into memory'.format(len(rows)))
-
-    # Hand the row off to the database processings/insertion part of the script
-    rows2db(sesh, rows, error_file, log, args.diag)
-
+    run(args)
```
jameshiebert: You can move this and the ``parse_args()`` call into the ``if __name__ == '__main__'`` conditional and get rid of the ``main()`` function altogether.

tests/test_wmb_download.py
```diff
@@ -0,0 +1,5 @@
+from crmprtd.wmb_dir.download import logging_setup
+
+def test_logging_setup():
+    log = logging_setup('mof.log', 'test@mail.com', 'INFO')
+    assert log != None
```
jameshiebert: Are there any stronger guarantees that you can make about the return value of ``logging_setup()``? What type is it? What information does it need to contain to be functional for the rest of the program?

crmprtd/wmb_dir/normalize.py
```diff
@@ -0,0 +1,77 @@
...
+def save_file(reader, cache_dir, data):
+    # save the downloaded file
+    fname_out = os.path.join(cache_dir, 'wmb_download' + datetime.strftime(datetime.now(), '%Y-%m-%dT%H-%M-%S') + '.csv')
+    with open(fname_out, 'w') as f_out:
+        copier = csv.DictWriter(f_out, fieldnames = reader.fieldnames)
+        copier.writeheader()
+        copier.writerows(data)
+
+
+def reader2data(reader):
```
jameshiebert: Did you write this? This could just as easily be replaced by the Python ``list`` builtin.

```python
james@basalt:~/code/git/crmprtd/crmprtd$ python3.5
Python 3.5.2 (default, Nov 23 2017, 16:37:01) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> def an_iterator():
...     yield 'foo'
...     yield 'bar'
...     yield 'blah'
...     yield 'stuff'
... 
>>> for i in an_iterator():
...     print(i)
... 
foo
bar
blah
stuff
>>> print(list(an_iterator()))
['foo', 'bar', 'blah', 'stuff']
```

tests/test_wmb_normalize.py
```diff
@@ -0,0 +1,6 @@
+from crmprtd.wmb_dir.normalize import reader2data
+
+def test_reader2data():
```
jameshiebert: I'd remove the ``reader2data()`` function and this test.

jameshiebert:
I think that covers all of my comments. Looks great!

PR #18, opened 2018-06-14
Formatting clean up

nikola-rados:
Using ```autopep8``` and ```flake8``` the python code in crmprtd has been formatted. Travis has been updated to check formatting going forward.  This is achieved by using ```pytest-flake8``` in concert with the other pytest plugins.

**Some Notes:**
- ```compat.py``` was removed as it was no longer necessary
- the ```# noqa``` inline comment was used to exclude some items in the testing directory which do not have to be PEP8 compliant
- added to the ```.flake8```config file is an exclusion for ```ec_data.py``` which does not contain python code


crmprtd/crmprtd_log.txt
```diff
@@ -0,0 +1,3967 @@
+2018-05-25 14:42:12,310:DEBUG:crmprtd.wamr - Processing observations
```
jameshiebert: ``crmprtd_log.txt`` should not be part of this (or any) commit.

crmprtd/ec.py
```diff
@@ -29,20 +30,25 @@ def makeurl(freq='daily', province='BC', language='e', time=datetime.utcnow()):
     """
     fmt = '%Y%m%d%H' if freq == 'hourly' else '%Y%m%d'
     freq = 'yesterday' if freq == 'daily' else freq
-    fname = '%s_%s_%s_%s.xml' % (freq, province.lower(), time.strftime(fmt), language)
-    return 'http://dd.weatheroffice.ec.gc.ca/observations/xml/%s/%s/%s' % (province.upper(), freq, fname)
+    fname = '%s_%s_%s_%s.xml' % (
```
jameshiebert: Can we change this to use Python's new style string formatting, while we're working on this?

crmprtd/moti.py
```diff
@@ -26,72 +26,92 @@
...
+def makeurl(report='7110', request='historic',
+            station=None, from_=None, to=None):
     '''
     Construct a URL for fetching the data file
     '''
     fmt = '%Y-%m-%d/%H'
     from_ = from_.strftime(fmt)
     to = to.strftime(fmt)
-    url  = 'https://prdoas2.apps.th.gov.bc.ca/saw-data/sawr{report}?request={request}&station={station}&from={from_}&to={to}'.format(**locals())
+    url = (('https://prdoas2.apps.th.gov.bc.ca/saw-data/sawr{report}?'
```
jameshiebert: Looks like you have changed the return type from a string to a tuple here. That's not what we want, is it?

jameshiebert: Or maybe that's a new multi-line string syntax that I don't know about?

crmprtd/crmprtd_log.txt
nikola-rados: You are right, I'm not sure how that got in there. I'll take it out.

crmprtd/moti.py
nikola-rados: I think I accidentally put an extra set of parentheses.  The multi line string format I've been following I found [here](https://stackoverflow.com/questions/10660435/pythonic-way-to-create-a-long-multi-line-string).

crmprtd/ec.py
nikola-rados: Yes, do you want it done just in this particular instance, or everywhere in the project?

jameshiebert: Just this instance I guess. Or... only in places where we're changing the code anyway. There shouldn't be too many of these left... we've been using str.format for a long time now.

nikola-rados: Ok, I'll have a look and see if I spot it anywhere else.

jameshiebert:
Looks great! I'll merge this into master, and please feel free to merge it into dev (and resolve any conflicts as necessary).

PR #23, opened 2018-07-09
Pipeline refactor for normalization

nikola-rados:
As part of the ongoing effort to refactor ```crmprtd``` to a pipeline, the [normalization](https://github.com/pacificclimate/crmprtd/blob/dev/crmprtd/__init__.py) stage has been added.  Since this pull request contains a rather large amount of changes I will try and detail some below.

### Changes to Other Modules
Since large parts of the code had to be reworked and re-purposed several other modules were touched.

**```[network name]/download.py```**

- the files previously contained an option to use a local file as input, this was removed as it was felt that it didn't fit within the scope of the "download" stage (please note that the command line parsing for the local files remains in the ```scripts/```, perhaps they should be removed if they are not being used going forward?)

- any logging configuration has been moved to the ```scripts/``` modules

- each module retrieves the network data a little bit differently, however the output for each has been changed to a [SpooledTemporaryFile](https://docs.python.org/2/library/tempfile.html#tempfile.SpooledTemporaryFile) with an environment variable dictating the amount of data saved to memory before moving to disk

- the output is yielded rather than returned

**```[network name]/__init__.py```**

- contains plenty of old code that is currently unused

- unused code has not been deleted as some of it could be re-purposed for other stages in the pipeline

**```scripts/```**

- previously called the ```run()``` method from the network specific download module but now controls the pipeline

- logging configuration has been moved here

### So what is new in normalize.py?
The ```[network name]/normalize.py``` takes an open file stream as input and ouputs a [namedTuple](https://docs.python.org/2/library/collections.html#collections.namedtuple).  The namedTuple is a row which contains the following data attributes:

- time (datetime)
- val (int)
- variable_name (str)
- unit (str)
- network_name (str)
- station_id (str)
- lat (int)
- lon (int)

Here is an example namedTuple output from ```scripts/wamr_hourly```:

**```Row(time=datetime.datetime(2018, 6, 11, 8, 0, tzinfo=<DstTzInfo 'Canada/Pacific' LMT-1 day, 15:48:00 STD>), val=0.47, variable_name='WSPD_VECT', unit='m/s', network_name='WAMR', station_id='M114009', lat=None, lon=None)```**

These attributes are parsed and/or converted where necessary to produce the desired data type.  Since each of the networks have a different data format not all of these attributes are able to be filled.  Where this occurs ```None``` is set as the value.  Furthermore, at this point all data is turned into a namedTuple, there are no checks made to see if the database accepts the variable or not.  Both of these things have been left out under the assumption the the "align" stage would handle these types of checks.


jameshiebert:
> Since each of the networks have a different data format not all of these attributes are able to be filled. Where this occurs None is set as the value.

I had forgotten that some of the networks only provide station location information "out of band" and not in the data stream. This breaks our model a *little* bit, but I like the way you've handled it. We'll just have to remember this in the align stage and have to choose and appropriate beavhiour (and user notification) when we can't necessarily verify that the stations match with incomplete information.

Full review pending...

crmprtd/wmb/normalize.py
```diff
@@ -1,70 +1,87 @@
...
-    fname_out = os.path.join(cache_dir,
-                             'wmb_download' +
-                             datetime.strftime(datetime.now(),
-                                               '%Y-%m-%dT%H-%M-%S') +
-                             '.csv')
-    with open(fname_out, 'w') as f_out:
-        copier = csv.DictWriter(f_out, fieldnames=reader.fieldnames)
-        copier.writeheader()
-        copier.writerows(data)
+        for key, value in d.items():
```
jameshiebert: So this isn't necessarily wrong, but it's a little weird to iterate over everything, but skip the first two things. But I get it. I can't think of an obviously cleaner way to do it off the top of my head. But I had to read the code carefully *and* look at sample input to understand what you were doing here. Maybe add an explanatory comment at the top of the block about out the input contains a single line for each hour of observation with multiple variable quantities.

nikola-rados:
@jameshiebert Changes you recommended have been made. Just some items to note:

- ec and moti download.py yields using the iter_lines() rather than iter_content()
- both ec and moti do not use the tee() function as the "file" can be written without being consumed
- wamr and wmb both still use ftp + SpooledTemporaryFile (should this be changed to a regular TemporaryFile?)
- wamr and wmb use the tee() if cmd line arguement given

Please let me know if you would like anything else done here.

scripts/hourly_wmb.py
```diff
@@ -77,5 +80,12 @@
...
     for file in download(args):
-        for row in normalize(file):
+
+        if args.cache_file:
+            to_cache, file_stream = tee(file)
+            with open(args.cache_file, 'w') as f:
+                for line in to_cache:
+                    f.write(line)
+
+        for row in normalize(file_stream):
```
jameshiebert: I *think* that this will fail, if `args.cache_file` is not defined. Because then `file_stream` will not be defined...

scripts/wamr_hourly.py
```diff
@@ -88,5 +89,12 @@
...
     for file in download(args):
-        for row in normalize(file):
+
+        if args.cache_file:
+            to_cache, file_stream = tee(file)
+            with open(args.cache_file, 'w') as f:
+                for line in to_cache:
+                    f.write(line)
+
+        for row in normalize(file_stream):
```
jameshiebert: Likewise.

scripts/hourly_wmb.py
nikola-rados: uh oh, yes it absolutely will, I'll change it such that file_stream will just be named file.

jameshiebert:
I've done a bit of refactoring to make the pipeline usage consistent across modules. With that done, it looks like we're all good to go!

PR #24, opened 2018-07-24
Feature Align

nikola-rados:
### Overview
The output of the _normalization phase_  is a namedTuple **Row**, this is then fed into the _align phase_ which produces a **pycds.Obs** object.  This object has 4 attributes: _time, vars_id, history_id, datum_.  Align ensures that the attributes found in **Row** are valid for our purposes and translates them into the **pycds.Obs** object.  In order for this object to be produced a place (location) and a thing (data) are required.  These 2 items have several checks to ensure that they are suitable for database entry.

### Place
Requirements for place are **network_name** and **station_id**.  Both of these attributes are used to determine the **history_id** of the station found in the _crmp.meta_history_ table. If there is no matching station currently in the database, new station and history entries will be made.

### Thing
Requirements for thing are **time** and **value**.  Value has an associated **variable name**, which must already exist in the database in order to be valid.  Once the variable name is confirmed the value's **units** must also match.  If the units do no match those found in the database they will be converted.  These checks for value help us find the correct **vars_id** and ensures that **datum** has the correct units.

### Flowchart
Below is the basic logic the code follows.
align phase


crmprtd/align.py
```diff
@@ -8,12 +8,237 @@
...
+                log.debug('Station set is empty, adding new station')
+                hid = create_station_and_history_entry(sesh, obs_tuple)
+
+            elif len(close_stns) == 1:
+                log.debug('Single station in set')
+                hid = close_stns.pop()
+
+            elif len(close_stns) > 1:
+                log.debug('Multiple stations in set')
+                closest_pos = 999   # max distance should be 800m at this point
```
jameshiebert: IIRC `closest_stns_within_thresholds` returns its result [*ordered by distance*](https://github.com/pacificclimate/crmprtd/blob/dev/crmprtd/ec/__init__.py#L159). So if you want to select the closest, you should just be able to select the first one.

crmprtd/align.py
```diff
@@ -8,12 +8,237 @@
...
+                if q.count() > 0:
+                    log.debug('Matched hid using sdate/edate')
+                    hid, = q.first()
+                    break
+
+        if not hid:
+            log.warning('Unable to match station',
+                        extra={'station_id': obs_tuple.station_id,
+                               'network_name': obs_tuple.network_name})
+            return
```
jameshiebert: What is the meaning of an empty return? Does the caller of `align()` expect and handle a `None` value being returned?

crmprtd/align.py
```diff
@@ -8,12 +8,237 @@
...
+
+    # thing
```
jameshiebert: So I know that *I* called this a "thing" in our introductory discussion about this, but it was simply because I couldn't think of a better way to describe it. We should be more specific here. "Quantity"? "Variable"? "Measurement"?

crmprtd/align.py
```diff
@@ -8,12 +8,237 @@
...
+                    'vars_id': obs_tuple.variable_name})
+        return
+
+    vars_id, = q.first()
+    log.debug('Observation variable matches', extra={'vars_id': vars_id})
+
+    log.debug('Check unit')
+    if obs_tuple.unit is None:
+        log.debug('No unit given, checking if table contains unit')
+        unit, = sesh.query(Variable.unit).join(Network).filter(and_(
```
jameshiebert: If you simply query for `Variable` above (as I described early on in this review), then you can simply use the expression `variable.unit` here instead of explicitly writing another query.

crmprtd/align.py
```diff
@@ -8,12 +8,237 @@
...
+            Network.name == obs_tuple.network_name,
+            Variable.name == obs_tuple.variable_name)).first()
+
+        if not unit:
+            log.warning('Table contains no unit for variable',
+                        extra={'var_name': obs_tuple.variable_name})
+            return
```
jameshiebert: Going back to the empty return, this function is a generator since you are yielding things at the bottom. What happens if you yield a few things, return None, then try to yield a few more things? You'll never get to the things after the return (and in some versions of Python, mixing returns and yields in a single function is a syntax error).

```python
$ python
Python 3.6.5 (default, May  3 2018, 10:08:28) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> def foo():
...     yield 1
...     yield 2
...     yield 3
...     return
...     yield 4
...     yield 5
... 
>>> for i in foo():
...     print(i)
... 
1
2
3
```

So you want to better define what happens in an error condition here. You *probably* want to log.error and then `continue` to the next iteration.

crmprtd/ec/__init__.py
```diff
@@ -369,7 +395,7 @@ def db_unit(sesh, var_name):
         return None
 
 
-class OmMember:
+class OmMember(object):
```
jameshiebert: Good catch.

crmprtd/moti/__init__.py
```diff
@@ -3,15 +3,32 @@
...
 from sqlalchemy.exc import IntegrityError
 
 from pycds import History, Network, Station, Variable, Obs
 from crmprtd import Timer
 
-
 log = logging.getLogger(__name__)
 
+
+def logging_setup(log_conf, log, error_email, log_level):
```
jameshiebert: Why has logging_setup been repropogated and duplicated again?

crmprtd/wamr/normalize.py
```diff
@@ -1,30 +1,57 @@
...
-def prepare(rows, error_file, log, args):
-    # Database connection
-    try:
-        engine = create_engine(args.connection_string)
-        Session = sessionmaker(engine)
-        sesh = Session()
-    except Exception as e:
-        log.critical('Error with Database connection', exc_info=True)
-        sys.exit(1)
+        # FIXME: This is a bit janky, does it belong in normalize?
```
jameshiebert: Yeah, I think it's fair to normalize the *names* of units in the normalize functions.

scripts/hourly_wmb.py
```diff
@@ -71,4 +76,15 @@
...
+
+    engine = create_engine(args.connection_string)
+    Session = sessionmaker(engine)
+    sesh = Session()
+    sesh.begin_nested()
+
+    # Pipeline
+    for file in download(args):
+        for row in normalize(file):
+            for obs in align(sesh, row):
```
jameshiebert: This isn't quite right. The way you have expressed this there are multiple files per download, and there multiple observations *per row* from the align phase. There should be only *one* obs per row. See [my pipeline PR](https://github.com/pacificclimate/crmprtd/pull/28) for how I think the pipeline should be.

tests/test_wmb_download.py
```diff
@@ -1,7 +1,7 @@
-from crmprtd.wmb.download import logging_setup
+from crmprtd.wmb import setup_logging
```
jameshiebert: Looks like this PR drops the test coverage from 62% down to 55% so presumably, there's a lot of untested code in here. Can you start putting together a set of tests for your align phase? That *may* help you to decompose it a bit more to think of it in testable units.

crmprtd/align.py
jameshiebert: **Update:** I forgot that the argument to this function is just a single Obs tuple and the result is a single Obs object. That being the case, it's probably appropriate to return None (i.e. an empty return statement), but then you'll want to `return` the Obs at the bottom of the function, rather than `yield`ing it. There's not really much point to a `yield` of a single object, since you're not going to iterate over a single thing.

nikola-rados:
@jameshiebert I have made changes based on your feedback.  The biggest change is the structure of align, everything has been broken down into bite sized pieces.  Hopefully this clears up the logic and makes the code more self-documenting.  Furthermore I have added testing for align and it is **100%** covered.  Let me know if you think I should break some of the tests up into separate cases (mainly the ```test_align_successes``` and ```test_align_failures```).  Just a note, ```closests_stns_within_threshold()``` was moved from ```crmprtd/ec/__init__.py``` into ```align.py``` as the addition of ```run_data_pipeline()``` caused a circular dependency.

Looking forward to your feedback on these changes.

crmprtd/align.py
```diff
@@ -8,12 +8,222 @@
...
+        Variable.name == variable_name)).first()
+
+    if not variable:
+        log.debug('Unable to match variable')
+        return None
+
+    return variable
+
+
+def get_history(sesh, obs_tuple):
```
jameshiebert: Good, I like this function. It clearly breaks up the various cases (no matches, an exact match, multiple matches) and delegates as necessary. It's clear what's happening.

One improvement that I would suggest is to only pass in the information that is necessary (i.e. the location related info) from the obs_tuple. It will lengthen the argument list, which isn't always desirable, but OTOH it will make it more clear what information the function needs. *And* many of the delegates need a different set of information, so it will make it easier to split up the info.

```
def get_history(sesh, network_name, native_id, lat, lon, elevation):
   ...
```
Would that cover it?

crmprtd/align.py
```diff
@@ -8,12 +8,222 @@
...
+    log.debug('Find history entry')
+    history = sesh.query(History).join(Station).join(Network).filter(and_(
```
jameshiebert: This could be multiple history entries. We should name the variable `histories`.

crmprtd/align.py
```diff
@@ -8,12 +8,222 @@
...
+        Network.name == obs_tuple.network_name,
+        Station.native_id == obs_tuple.station_id))
+
+    if history.count() == 0:
+        return create_station_and_history_entry(sesh, obs_tuple)
+    elif history.count() == 1:
+        return history.first()
+    elif history.count() >= 2:
+        return match_station(sesh, obs_tuple, history)
```
jameshiebert: Shouldn't the parameter be `histories.all()`?

crmprtd/align.py
```diff
@@ -8,12 +8,222 @@
...
+
+    for id in close_stns:
+        for history in histories:
+            if id == history.id:
+                log.debug('Matched history',
+                          extra={'station_name': history.station_name})
+                return history
+
+
+def match_station(sesh, obs_tuple, history):
```
jameshiebert: Since this is only called from the `>=2` branch of the conditional, we should assume that it is being called with multiple history objects. So rename `history` to `histories` and keep the number consistent across the code base.

crmprtd/align.py
```diff
@@ -8,12 +8,222 @@
...
+                      extra={'station_name': hist.first().station_name})
+            return hist.first()
+
+    log.warning('Could not find match',
+                extra={'network_name': obs_tuple.network_name,
+                       'native_id': obs_tuple.station_id})
+    return None
+
+
+def match_station_with_location(sesh, obs_tuple, histories):
```
jameshiebert: Can we name this `find_nearest_history`? Also, I just noticed that `closest_stns_within_threshold` doesn't take a *network* parameter. That's *probably* a bug... well not strictly a bug, but a mistake that may or may not impact us depending on the data. It would probably be wise to add a network_name parameter to that function and add that as a filter.

crmprtd/align.py
```diff
@@ -8,12 +8,222 @@
...
+    log.debug('Check if network in db', extra={'network_name': network_name})
+    network = sesh.query(Network).filter(
+        Network.name == network_name)
+    return network.count() != 0
+
+
+def align(sesh, obs_tuple):
+    log.info('Begin alignment')
+    # Without these items an Obs object cannot be produced
+    if obs_tuple.network_name is None or obs_tuple.time is None or \
```
jameshiebert: Good check here. While it's only a single line, I'd be tempted to split this out to a separate function named `has_required_information(obs_tuple)`. Then the code becomes self-documenting and the comment becomes irrelevant.

crmprtd/align.py
```diff
@@ -8,12 +8,222 @@
...
+            obs_tuple.val is None or obs_tuple.variable_name is None:
+        log.warning('Observation missing critical information',
+                    extra={'network_name': obs_tuple.network_name,
+                           'time': obs_tuple.time,
+                           'val': obs_tuple.val,
+                           'variable_name': obs_tuple.variable_name})
+        return None
+
+    if not is_network(sesh, obs_tuple.network_name):
+        log.warning('Network does not exist in db',
```
jameshiebert: I'd call this an error.

crmprtd/align.py
```diff
@@ -8,12 +8,222 @@
...
+
+def is_network(sesh, network_name):
+    log.debug('Check if network in db', extra={'network_name': network_name})
+    network = sesh.query(Network).filter(
+        Network.name == network_name)
+    return network.count() != 0
+
+
+def align(sesh, obs_tuple):
+    log.info('Begin alignment')
```
jameshiebert: If you're calling this for every observation, `log.info()` is going to be extremely noisy. I'd remove this.

crmprtd/align.py
```diff
@@ -8,12 +8,222 @@
...
+
+    datum = unit_check(sesh, obs_tuple.unit, variable.unit, obs_tuple.val)
+    if not datum:
+        log.warning('Unable to confirm data units',
+                    extra={'unit_obs': obs_tuple.unit,
+                           'unit_db': variable.unit,
+                           'data': obs_tuple.val})
+        return None
+
+    log.info('Completed align')
```
jameshiebert: Likewise, re: noisy.

crmprtd/align.py
```diff
@@ -8,12 +8,222 @@
...
+    return Obs(history=history,
```
jameshiebert: Sweet!

crmprtd/ec/__init__.py
```diff
@@ -9,6 +9,7 @@
 from sqlalchemy.sql import func
 from sqlalchemy.exc import IntegrityError
 from crmprtd import Timer
+from crmprtd.align import closest_stns_within_threshold
```
jameshiebert: This only gets used in code that you've obsoleted in this PR. I'd remove this line and the obsolete code.

tests/test_align.py
```diff
@@ -0,0 +1,250 @@
+import pytest
```
jameshiebert: Please allow me to wait to comment on your testing code until the next iteration of feedback. I think there are a few changes to the function definitions, so I'll provide detailed feedback on the testing once those are solidified.

crmprtd/align.py
```diff
@@ -8,12 +8,222 @@
...
+        return unit_db_check(unit_obs, unit_db, val)
+
+
+def match_station_with_active(sesh, obs_tuple, histories):
+    log.debug('Search for active stations')
+    for history in histories:
+        hist = sesh.query(History).filter(
+            and_(History.id == history.id,
+                 History.sdate != None,
+                 History.edate == None))    # noqa
```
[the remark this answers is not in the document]
nikola-rados: The behaviour of the query was different when I used ```is not``` and ```is``` rather than ```!=``` and ```==```.  I found an [answer](https://stackoverflow.com/questions/5602918/select-null-values-in-sqlalchemy) as to why this was happening.  However, ```flake8``` will not pass while using ```is not``` and ```is``` and rather than ignore this PEP8 violation altogether, I chose to ignore it for this case.

crmprtd/align.py
nikola-rados: I was thinking of doing this exact thing but decided against it last minute.  I'll add it in.

tests/test_align.py
nikola-rados: Sounds good! Thank you again for the feedback.

crmprtd/align.py
nikola-rados: When you say ```elevation``` do you mean the threshold used in ```closest_stns_within_threshold()```?

jameshiebert: No, the vertical elevation of the station. Maybe we don't get that from any of the networks?

nikola-rados: I haven't see that from any of them.

jameshiebert: OK, ignore that then. Make sure that there's a big fat warning in the station creation code though, because a station without an elevation is problematic for any further data analysis (particularly because we use elevation-based regression for interpolation).

nikola-rados: Sounds good, I'll add that in.

nikola-rados:
@jameshiebert I have addressed the feedback you have given.  Furthermore, _most_ of the code that has become obsolete has been removed as well as the accompanying tests.  I've written some simple tests for the normalization scripts to get the total coverage to 77%.  This percentage does not take the download scripts into consideration as they have been omitted.  Please let me know if there are any more changes you would like me to make before we move onto the testing feedback.

crmprtd/__init__.py
```diff
@@ -167,7 +169,12 @@ def run_data_pipeline(download_func, normalize_func, download_args):
...
-    for row in rows:
-        print(row)
-    # observations = [align(row) for row in rows]
+
+    engine = create_engine(args.connection_string)
+    Session = sessionmaker(engine)
+    sesh = Session()
+
+    observations = [align(sesh, row) for row in rows]
+    for ob in observations:
```
jameshiebert: Want to add that filter for `None` values here that we discussed?

nikola-rados: Oops, forgot to add that in. My mistake.

jameshiebert:
W00t!

PR #27, opened 2018-07-26
Refactor Argparse

nikola-rados:
The four main scripts contained plenty of the same ArgParser arguments.  Any common arguments were put into a method in ```crmprtd/__init__.py```.  Some of the single letter identifiers were changed to prevent duplicates between the common and unique args.

A side effect of these changes was the behavior of **--log_conf** argument.  It was common to the scripts but was not quite implemented the same everywhere.  In order to standardize it across the modules the **setup_logging** method was changed and subsequently moved into ```crmprtd/__init__.py```.  Through the changes it had become duplicate code which could be factored out.


crmprtd/__init__.py
```diff
@@ -65,6 +68,54 @@ def __exit__(self, *args):
         self.run_time = self.end - self.start
 
 
+def common_script_arguments(parser):
```
jameshiebert: Can we exclude this function from our test coverage calculation? This is all standard library code and I don't think that there's value to testing it. Or just write a simple test function that passes it a parser, and checks that it gets a parser object back...

nikola-rados:
Those changes were made and we're ready to go!

jameshiebert:
Perfect. LGTM!

PR #30, opened 2018-08-08
Add option for logging configuration

nikola-rados:
The changes are rather minor, but the ```--log_conf``` argparse argument was going unused in some of the scripts.  The ```setup_logging``` method has been moved to common ground and is shared by all the scripts.  Furthermore the default for ```--log_conf``` was changed to ```None``` to make it easier identify when the default ```resource_stream``` should be used.

A further point is that I discovered this issue exists in the ```dev``` branch as well.  I'm unsure if it is better to attempt a merge of this branch into ```dev```, or just handle it in a separate fix.


crmprtd/__init__.py
```diff
@@ -51,3 +55,26 @@ def f_retry(*args, **kwargs):
             return f(*args, **kwargs)
         return f_retry  # true decorator
     return deco_retry
+
+
+def setup_logging(log_conf, log, error_email, log_level, name):
+    if log_conf:
+        with open(log_conf, 'rb') as f:
+            log_c = yaml.load(f)
```
jameshiebert: Let's change the name of `log_c` to be `base_config`. So we start with a base config and then optionally modify it.

crmprtd/__init__.py
```diff
@@ -51,3 +55,26 @@ def f_retry(*args, **kwargs):
             return f(*args, **kwargs)
         return f_retry  # true decorator
     return deco_retry
+
+
+def setup_logging(log_conf, log, error_email, log_level, name):
```
jameshiebert: Let's change the name of the `log` argument to be `log_filename`, since that's what it actually is.

crmprtd/__init__.py
```diff
@@ -51,3 +55,26 @@ def f_retry(*args, **kwargs):
...
+    if log_conf:
+        with open(log_conf, 'rb') as f:
+            log_c = yaml.load(f)
+    else:
+        log_c = yaml.load(resource_stream('crmprtd', '/data/logging.yaml'))
+
+    if log:
+        log_c['handlers']['file']['filename'] = log
+    else:
+        log = log_c['handlers']['file']['filename']
```
jameshiebert: I know that this is what the code used to do, but is there any value to this assignment? We just throw out the `log` variable a few lines later. Let's axe the `else` branch of this conditional.

nikola-rados:
I'm working on moving some of the work done in ```dev``` over to this branch, shouldn't be too difficult to take what we did there and plug it in.  As for your second point I see what you are saying.  As far as I could tell all the logging occurring in this modules is of the format specified by the config.  Also, I think we don't need to return a log object like we are currently.  Like you mentioned it isn't getting passed down like the modules used to do.  I will investigate further though.

jameshiebert:
> As far as I could tell all the logging occurring in this modules is of the format specified by the config.

OK cool. Just making sure... my ignorance of the Python's logging system just led me to "cautious skepticism" :)

jameshiebert:
From [here](https://docs.python.org/3/library/logging.html#logger-objects):

>Loggers that are further down in the hierarchical list are children of loggers higher up in the list. For example, given a logger with a name of foo, loggers with names of foo.bar, foo.bar.baz, and foo.bam are all descendants of foo.

So yeah, we should be good.

nikola-rados:
Awesome!

PR #32, opened 2018-08-14
Feature Insert

nikola-rados:
### Overview
The final stage of the crmprtd pipeline has been added.  The insertion stage takes a list of ```pycds Obs()``` objects and inserts them into a database.

### What it does
Once the ```pycds Obs()``` list is passed in it is checked for duplicate observations.  This is achieved by taking samples from the list and doing individual checks.  If all of the samples are duplicates, a one-by-one insertion strategy is used.  It was found that in worst-case scenarios (all observations are duplicates) a one-by-one insertion strategy was consistently faster than the bisecting strategy.  Furthmore, if the samples come back and they are unique, the bisecting insertion strategy is used.  During testing the bisecting strategy proved to be faster than one-by-one in all cases but the worst case.  The number of successful and unsuccessful observations are tracked, as well as the duration of the insertion process. These results are returned from the module.

### File Changes
Some files have been renamed or deleted during this step.
 - ```db.py```  was deleted and its contents were moved to ```insert.py```
 - ```test_db.py``` was renamed ```test_insert.py```
 - ```wmb_exceptions.py``` was renamed ```db_exceptions```

### Testing
Testing has been added for ```insert.py```.  The ```test_db.py``` code that existed before has been altered slightly and new tests have been added.  Coverage for the ```insert.py``` module is currently **83%**.

### Speed Testing
A new directory and files have been added for speed testing.  The files are not a part of the build package and are excluded from the coverage testing.  The files have been included for reference but are not expected to be used going forward.

### Argparse Additions
Two new args were added to the ```common_script_args```, chunk_size and sample_size.  These apply to the insertion strategy and sampling respectively (both have defaults).


crmprtd/__init__.py
```diff
@@ -101,6 +91,14 @@ def common_script_arguments(parser):    # pragma: no cover
                               'observations')
     parser.add_argument('-i', '--input_file',
                         help='Input file to process')
+    parser.add_argument('--chunk_size', type=int,
+                        default=4096,
```
jameshiebert: Ahhhh, so I finally had an epiphany about why the chunking+bisecting is faster than bisecting alone and that's because you use a power of two for the base chunk size. I *bet* that we can get comparable performance without the need for configuration by breaking the observation set into chunks comprised of maximal powers of two. E.g. if you have 1200 observations, break it up into (1024, 128, 32, 16). I think we're running out of time, so we should relegate this to future work, but I *bet* that it would be the fastest.

crmprtd/__init__.py
```diff
@@ -189,11 +195,7 @@ def run_data_pipeline(download_func, normalize_func, download_args,
...
 
-    observations = [align(sesh, row) for row in rows if row]
-    for ob in observations:
-        print(ob)
-    # insert(observations)
-
-
-def subset_dict(a_dict, keys_wanted):
-    return {key: a_dict[key] for key in keys_wanted if key in a_dict}
+    observations = list(filter(lambda ob: ob is not None,
```
jameshiebert: I could be wrong, but I think that current practice is to shy away from using `filter` + `lambda`. I think that this can still be expressed with:

```
obs = [align(sesh, row) for row in rows]
obs = [ob for ob in obs if ob]
```

crmprtd/__init__.py
```diff
@@ -189,11 +195,7 @@ def run_data_pipeline(download_func, normalize_func, download_args,
...
+                               [align(sesh, row) for row in rows]))
+    results = insert(sesh, observations, **insert_args)
+    print(results)
```
jameshiebert: Do we want to log them too?

crmprtd/align.py
```diff
@@ -47,7 +47,7 @@ def closest_stns_within_threshold(sesh, network_name, lon, lat, threshold):
 
 
 def convert_unit(val, src_unit, dst_unit):
-    if src_unit != dst_unit:
+    if src_unit != dst_unit and src_unit is not None:
```
jameshiebert: What happens if `src_unit` *is* `None`? Or `dest_unit`? Seems to me that those should be error conditions (though I say so without looking at the calling context, so I'm happy to reconsider).

crmprtd/align.py
```diff
@@ -61,10 +61,11 @@ def convert_unit(val, src_unit, dst_unit):
 
 
 def unit_check(val, unit_obs, unit_db):
-    if not unit_obs and not unit_db:
+    if unit_obs is None and unit_db is None:
```
jameshiebert: Ah, the calling context! Should you `or` these together?

crmprtd/align.py
```diff
@@ -61,10 +61,11 @@ def convert_unit(val, src_unit, dst_unit):
...
         return None
     else:
-        return convert_unit(val, unit_obs, unit_db)
+        val = convert_unit(val, unit_obs, unit_db)
```
jameshiebert: Why do we need a temp var here? Can't you just return it?

crmprtd/align.py
```diff
@@ -114,17 +115,15 @@ def create_station_and_history_entry(sesh, network_name, native_id, lat, lon):
 
     network = network.first()
     stn = Station(native_id=native_id, network=network)
-    with sesh.begin_nested():
-        sesh.add(stn)
+    sesh.add(stn)
```
jameshiebert: Tell me more about why you removed the nested transaction here and below...

crmprtd/insert.py
```diff
@@ -8,11 +8,217 @@
...
+        return self
+
+    def __exit__(self, *args):
+        self.end = time.time()
+        self.run_time = self.end - self.start
+
+
+def chunks(list, chunk_size):
+    for i in range(0, len(list), chunk_size):
+        yield list[i:i+chunk_size]
```
jameshiebert: Hmmm interesting. This does work, but only because Python will (surprisingly) drop the end of an array if you give it an index that is larger than the length of the array.

I suspect that you could do this more elegantly with `itertools.groupby`, but this works, so I'm satisfied to go with this.

crmprtd/insert.py
```diff
@@ -8,11 +8,217 @@
...
+
+
+def get_sample_indices(num_obs, sample_size):
+    if sample_size > num_obs:
```
jameshiebert: Good safety check :)

crmprtd/insert.py
```diff
@@ -8,11 +8,217 @@
...
+    sample_indices = [0]
+    index = 0
+    for i in range(sample_size - 1):
+        index += index_buffer
+        sample_indices.append(index)
+
+    return sample_indices
+
+
+def is_unique(sesh, history_id, vars_id, time):
```
jameshiebert: Perfect. Although, I'd maybe rename the function to `obs_exists`.

crmprtd/insert.py
```diff
@@ -8,11 +8,217 @@
...
+    q = sesh.query(Obs.id).filter(
+        and_(Obs.history_id == history_id, Obs.vars_id == vars_id,
+             Obs.time == time))
+    if q.count() > 0:
+        return False
+
+    return True
```
jameshiebert: It's logically the same, but I'd probably say:

```
    else:
        return True
```

There are only two options, so it's maybe slightly more clear to be explicit that either one outcome or the other happens.

crmprtd/insert.py
```diff
@@ -8,11 +8,217 @@
...
+        self.end = time.time()
+        self.run_time = self.end - self.start
+
+
+def chunks(list, chunk_size):
+    for i in range(0, len(list), chunk_size):
+        yield list[i:i+chunk_size]
+
+
+def get_sample_indices(num_obs, sample_size):
```
jameshiebert: I think that you could actually replace this whole function with:

```
    return random.sample(range(num_obs), sample_size)
```

crmprtd/insert.py
```diff
@@ -8,11 +8,217 @@
...
+    sample_obs = []
+    sample_indices = get_sample_indices(len(observations), sample_size)
+
+    for index in sample_indices:
+        sample_obs.append(observations[index])
+
+    already_exists = 0
+    for sample in sample_obs:
+        if not is_unique(sesh, sample.history_id, sample.vars_id, sample.time):
+            already_exists += 1
```
jameshiebert: Rather than having a counter, and returning False if all of the pass, you can do the opposite: return True if the is_unique conditional passes. Then you don't need the counter. E.g.:

```python
for sample in sample_obs:
    if is_unique(...):
        return True
return False
```

crmprtd/insert.py
```diff
@@ -8,11 +8,217 @@
...
+
+
+def single_insert_obs(sesh, o, dbm):
+    # Check to see if this entry will be unique
+    try:
+        if not is_unique(sesh, o.history_id, o.vars_id, o.time):
+            log.warning("Failure, observation already exists.")
+            dbm.failures += 1
+            raise UniquenessError(o)
+    except UniquenessError as e:
```
jameshiebert: You don't need the "exception handling" here, because you just threw the exception yourself. Simply use the conditional:

```
if not is_unique(...):
    log.warning(...)
    dbm.failures += 1
```

What are the cases in which the generic Exception would be triggered? There's not much to go wrong in the try branch...

crmprtd/insert.py
```diff
@@ -8,11 +8,217 @@
...
+                             vars_id=o.vars_id, hid=o.history_id, e=e)
+
+    # value does not exist in obs_raw, continue with insertion
+    try:
+        sesh.add(o)
+        sesh.commit()
+        log.debug("Successfully inserted observation")
+        dbm.successes += 1
+        return 1
+    except Exception as e:
```
jameshiebert: This *is* appropriate exception handling, because an error isn't something that we "expect" I.e. it's *exceptional* behaviour.

crmprtd/insert.py
```diff
@@ -8,11 +8,217 @@
...
+            log.info("Successfully inserted observations",
+                     extra={'num_obs': len(obs)})
+            sesh.commit()
+            dbm.successes += len(obs)
+            return len(obs)
+    return 0
+
+
+def insert(sesh, observations, chunk_size, sample_size):
+    dbm = DBMetrics()
```
jameshiebert: I like the use of the metrics object here. I hadn't really thought through how we were going to do this, and that's a good solution.

crmprtd/insert.py
```diff
@@ -8,11 +8,217 @@
...
+
+    if has_unique_obs(sesh, observations, sample_size):
+        log.info("Using Single Insert Strategy")
+        with Timer() as tmr:
+            for ob in observations:
+                try:
+                    single_insert_obs(sesh, ob, dbm)
+                except Exception:
+                    pass    # FIXME: Better way to handle exception here?
```
jameshiebert: Oooooh yeah, you gotta do *something* here. passing in the exception handler is worse than simply omitting the try/except block.

crmprtd/align.py
nikola-rados: I was looking for an issue with the insertion and thought that perhaps I was using the nested transactions incorrectly.  I actually meant to change this back.  Just a question, does using the ```sesh.begin_nested()``` as a context manager commit upon ```__exit__()```? Or is the commit at the bottom of the method still neccessary?

jameshiebert: @rod-glover Can you comment on this?

crmprtd/insert.py
nikola-rados: The reason for the counter is to see if **all** samples are duplicates.  The return value is used to determine whether to use the one-by-one or chunks+bisecting.  Do you want to keep that condition? Or if there are _any_ duplicates found in the samples we use the one-by-one strategy?

jameshiebert: Right, but logically for a given array of booleans `A`: `any(not(A)) == not(all(A))`. Maybe I didn't give you the right logic in my first comment, but my point is that if you get the logic right, it can be done w/o a counter.

jameshiebert: I think the name of the function is confusing me. What is the expected return value. True if it detects all duplicates and False otherwise? If so maybe the name should be `contains_all_duplicates`? Then I guess the logic would be:

```
for sample in sample_obs:
    if is_unique(...):
        return False
return True
```
?

nikola-rados: I think you are right.  I'll change some of the names to make things easier to understand.  I'm looking a it a day after I wrote it and I'm having trouble reading through it.

nikola-rados:
@jameshiebert I have made the requested changes.  I've also included a new chunk method that will split the observations into power of 2 chunks.  The output of the ```align.py``` module has been slightly altered.  This was to circumvent the cascading backref issue you found.  Also in this push the rest of the unused code has been removed along with the accompanying tests.  Looking forward to your feedback.

crmprtd/insert.py
```diff
@@ -8,11 +8,210 @@
...
+    def __exit__(self, *args):
+        self.end = time.time()
+        self.run_time = self.end - self.start
+
+
+def pow_two_chunk(num):
+    return 2 ** floor(mathlog(num, 2))
+
+
+def get_chunk_sizes(remainder):
```
jameshiebert: Awesome!

jameshiebert:
There are two minor things that I would change if we had more time:
* I think the bisect insert strategy and the one-by-one insert strategy should use the same calling style. I.e. you given them each a list of Obs and they figure out what to do with them. Rather than what you have written, where single insert only takes *one* Obs and the caller has to iterate. Not a huge deal, but changing that would make the two functions consistent
* I like the use of the ``DBMetrics`` container to track successes/failures, but (a) I think that we need a "skips" metric (uniqueness isn't really a *failure* per se) and (b) delegate functions shouldn't modify the container quietly inside their function. Better to be explicit, where the delegate functions *return* the container. Otherwise, we're counting on side effects for critical functionality.

I don't think that we have time for these now, so I'll log an issue and we can circle back to them if we have time.

PR #35, opened 2018-08-23
Feature input file

nikola-rados:
The changes in this branch restructure how the pipeline is run internally.  ```download.py``` is now called directly and its output can be piped into either a file, or the new ```process.py``` module.  This module handles the final 3 stages in the pipeline.  

### Notes
- Output is pickled between ```download.py``` and ```process.py```.  The purpose of this was to ensure that the object type was preserved during its trip through stdout.  ```wamr```  and ```wmb``` both output strings, however ```ec``` and ```moti``` output bytes.
- There is some overlapping code that I left in for reference until changes are finished.  This change effectively negates the need for any of the main scripts.

### DIFF
--- a/crmprtd/ec/download.py
+++ b/crmprtd/ec/download.py
@@ -4,7 +4,6 @@ import logging
 import logging.config
 from datetime import datetime, timedelta
 from argparse import ArgumentParser
-import pickle
 
 # Installed libraries
 import requests
@@ -47,9 +46,8 @@ def download(time, frequency, province, language):
             raise IOError(
                 "HTTP {} error for {}".format(req.status_code, req.url))
 
-        pickle.dump('Network module name: ec', sys.stdout.buffer)
         for line in req.iter_content(chunk_size=None):
-            pickle.dump(line, sys.stdout.buffer)
+            sys.stdout.buffer.write(line)
 
     except IOError:
         log.exception("Unable to download or open xml data")
--- a/crmprtd/ec/download.py
+++ b/crmprtd/ec/download.py
@@ -3,12 +3,15 @@ import sys
 import logging
 import logging.config
 from datetime import datetime, timedelta
+from argparse import ArgumentParser
+import pickle
 
 # Installed libraries
 import requests
 
 # Local
 from crmprtd.ec import makeurl
+from crmprtd import common_script_arguments, setup_logging, logging_args
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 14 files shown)


nikola-rados:
@jameshiebert  the pickling has been removed in favour of your far simpler approach.  The output of ```[network_name]/download.py``` will always be bytes to stdout.  The two networks that use strings for their inputs (```wamr, wmb```) have a ```.decode()``` in normalization now, and their tests have been changed to reflect this change.  Furthermore, a new arguement was added '-N' or '--network' when you can specify which network the data is coming from. 

EX. ```python process.py -N wamr -c 'some_connection_string' < wamr_download.txt```

crmprtd/process.py
```diff
@@ -0,0 +1,78 @@
...
+    # FIXME: I do not think this arg gets used anywhere
+    parser.add_argument('-D', '--diag',
+                        default=False, action="store_true",
+                        help="Turn on diagnostic mode (no commits)")
+    parser.add_argument('--sample_size', type=int,
+                        default=50,
+                        help='Number of samples to be taken from observations '
+                             'when searching for duplicates '
+                             'to determine which insertion strategy to use')
+    parser.add_argument('-N', '--network',
```
jameshiebert: We should add the "choices" argument here, so the user knows what networks are available. For example *I* can't even remember whether I need to say `-N ec` or `-N EC_raw`, so guidance would be helpful.

crmprtd/ec/download.py
```diff
@@ -43,8 +45,38 @@ def download(time, frequency, province, language):
...
+                        help="'e' (english) | 'f' (french)")
+    parser.add_argument('-F', '--frequency', required=True,
+                        choices=['daily', 'hourly'],
+                        help='daily|hourly')
+    parser.add_argument('-t', '--time',
+                        help=("Alternate *UTC* time to use for downloading "
+                              "(interpreted using "
+                              "format=YYYY/MM/DD HH:MM:SS)"))
+    parser.add_argument('-T', '--threshold', default=1000,
+                        help=('Distance threshold to use when matching '
```
jameshiebert: We should specify the units (meters) here.

crmprtd/process.py
```diff
@@ -0,0 +1,78 @@
...
+def get_data():
+    for line in sys.stdin.buffer.readlines():
+        yield line
+
+
+def get_normalization_module(network):
+    return import_module('crmprtd.{}.normalize'.format(network))
+
+
+def process(connection_string, sample_size, network):
```
jameshiebert: This is beautiful code. Great work!

PR #36, opened 2018-08-23
Resolves #34

### LINKED ISSUE #33: Improve DBMetrics usage
The ``DBMetrics`` class in `insert.py` needs to have a third category for "skips" (observations that it has been asked to insert, but already exist in the database).

Additionally, the code should be refactored so that ``DBMetrics`` objects cannot be modified in delegate functions. Results should be loudly and explicitly returned, not quietly modified.

### LINKED ISSUE #34: Make insert usage consistent in insert.py
We have two insertion strategies in `insert.py`: bisect and one-by-one. Both have slightly different usages: the former takes a series of `Obs` objects, while the latter only takes one. Make these consistent. Move the loop into single_insert_obs, handle the uniqueness checks *inside* the function, and get rid of any usage of the `UniquenessError` exception which we don't need (it's expected behaviour, which we can easily test for, so it's not exceptional).

nikola-rados:
Changed the calling of ```single_insert_obs``` to be similar to ```bisect_insert_strategy```.  Also removed the use of ```UniquenessError`` and instead just increment ```dbm.failures``` and continue.  Tests were updated to reflect these changes.

**Edit**: I referenced the wrong issue in my commit and in the title of this PR, was #33 now #34

### DIFF
--- a/crmprtd/db_exceptions.py
+++ b/crmprtd/db_exceptions.py
@@ -19,16 +19,3 @@ class InsertionError(Error):
         for k, v in self.data.iteritems():
             s += '{0}: {1}, '.format(k, v)
         return s
-
-
-class UniquenessError(Error):
-    """
-    Exception raised when a tuple already exists in the database
-    """
-
-    def __init__(self, obs_id=None):
-        self.obs_id = obs_id
-
-    def __str__(self):
-        return ('This tuple already exists in database with obs_id '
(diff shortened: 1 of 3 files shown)


jameshiebert:
FYI, next time you typo the commit message, you can do a "git commit --amend" to fix it (and if you've already pushed, follow up with a "git push -f" assuming no one else has pulled the branch yet).

PR #41, opened 2018-08-27
 Replace scripts with console_scripts entry point 

jameshiebert:
This branch removes all of the individual networks' scripts in favor of a console_script pointer to each network's download function and a single console_script pointer to `process()` function that can handle the remaining three stages of the pipeline.


jameshiebert:
@nikola-rados I'm pretty keen to merge this, but I wanted to give you the change to look it over first and make sure that you understand it and are comfortable with the changes.

PR #46, opened 2019-11-21
Jenkins support

nikola-rados:
Add a `Jenkinsfile` to control the pipeline which:
- runs `py.test` in `python 3.5`, `3.6` and `3.7`
- pushes to `pypi` (will skip if packages already exist)
- cleans up the workspace

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,19 +1,74 @@
-node {
+/**
+ * Factored out common process for running test suite in python container
+ *
+ * @param pyimage Custom python image
+ */
+def test_with_container(String pyimage) {
     withDockerServer([uri: PCIC_DOCKER]) {
-        // Use image with gdal already installed
-        def pyenv = docker.image('python:3.5')
+        def pytainer = docker.image(pyimage)
 
-        pyenv.inside('-u root') {
-            stage('Install Requirements') {
-                sh 'pip install -r requirements.txt -r test_requirements.txt -i http://tools.pacificclimate.org/pypiserver/ --trusted-host tools.pacificclimate.org'
-                sh 'pip install .'
-                sh 'apt-get install postgresql postgis'
-            }
+        pytainer.inside() {
+            // sanity checks
+            sh 'which python'
+            sh 'python --version'
 
-            stage('Test Suite') {
-                sh 'py.test -v tests'
-            }
+
+            // try using pyenv plugin
+            // withPythonEnv('/usr/local/bin/python') {
+            //     sh 'pip install -r requirements.txt -r test_requirements.txt -i http://tools.pacificclimate.org/pypiserver/ --trusted-host tools.pacificclimate.org'
+            //     sh 'pip install .'
+            //     sh 'py.test -v --tb=short --cov --flake8'
+            // }
+
+            // // check for virtual env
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,71 @@
...
+        },
+        "Python 3.6": {
+            test_with_container('crmprtd-python36')
+        },
+        "Python 3.7": {
+            test_with_container('crmprtd-python37')
+        }
+    }
+
+    if (BRANCH_NAME == 'master') {
```
jameshiebert: Do we want to push to pypi for *every* master commit, or just for tagged releases? Is there a way to check for a tag here?

Jenkinsfile
```diff
@@ -0,0 +1,71 @@
...
+def push_to_pypi() {
+    withDockerServer([uri: PCIC_DOCKER]) {
+        def pytainer = docker.image('crmprtd-python35')
+
+        pytainer.inside {
+            // get twine
+            sh 'pip install twine'
+
+            // Build
+            sh 'python setup.py sdist'
```
jameshiebert: We should build a wheel too. So:

```
sh 'pip install twine wheel'

sh 'python setup.py sdist bdist_wheel'
```

You're `twine upload`ing everything in the dist/ directory so it should do both of them.

Jenkinsfile
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * @param pyimage Custom python image
+ */
+def test_with_container(String pyimage) {
+    withDockerServer([uri: PCIC_DOCKER]) {
+        def pytainer = docker.image(pyimage)
+
+        pytainer.inside {
+            // Installs
+            sh 'pip install -r requirements.txt -r test_requirements.txt -i http://tools.pacificclimate.org/pypiserver/ --trusted-host tools.pacificclimate.org'
```
jameshiebert: 1) Use https and you won't need the `--trusted-host` argument.
2) Cleaner to set the `PIP_INDEX_URL` environment variable than to use `-i`.

Jenkinsfile
nikola-rados: Sounds good, didn't notice I missed it!

Jenkinsfile
nikola-rados: Always happy to make it cleaner!

Jenkinsfile
nikola-rados: So this is something that I think will take some more work.  From what I've gathered Jenkins doesn't have a great way to do this and the alternative options I've tried have been unsuccessful.  I can take another swing at it as this is something I would want for other pipelines as well.

jameshiebert: Yeah, that'd be good. Thanks. What has been the challenge with it so far?

nikola-rados: The issue was that I could not get any information about the tags while in Jenkins.  Eventually I found that this was because the default source control checkout runs `git fetch` with the `--no-tags` flag. Once I figured that out checking for a tag was not difficult.

nikola-rados:
@jameshiebert I have addressed all of your comments, if you have new ones please let me know!

jameshiebert:
Awesome, I never knew about `git tag --contains`. That looks perfect. Merge away!

Issue #38, opened 2018-08-23
Add functionality for --diag


nikola-rados:
In the current data pipeline the ```--diag``` flag is not being used.  The purpose of this flag is to run the pipeline in diagnostic mode.  It should be passed into the ```align```,  ```insert``` since they interact with the database.

jameshiebert:
Ideally, this diagnostic mode could be implemented as a [context manager](https://docs.python.org/3/library/stdtypes.html#typecontextmanager) that opens a new transaction before align and rolls back the transaction after insert (if in the appropriate mode).

PR #48, opened 2019-11-26
Add diagnostic (no commit) mode to process script

jameshiebert:
This script adds a "diagnostic mode" to the process script. The mode was previously unused, so these commits add the code required to skip database modifications.

Basically when in use it disables the creation of any stations and skips over the creation of any new observations (instead logging them at level INFO).


jameshiebert:
Resolves #38 without the context manager (you can't commit/rollback if you use a read-only database session).

PR #51, opened 2020-01-03
Ignore MSNG values

jameshiebert:
We're seeing upwards of 100,000 log entries under level `ERROR` per day that are simply "observations" with value "MSNG" i.e. missing. We don't record missing values. If there's no data, we don't record it.

This patch demotes these events to just be ignored, or optionally logged at `DEBUG` level.


jameshiebert:
@nikola-rados and @faronium I think that this is pretty straightforward, but it doesn't hurt to have a couple eyes on it. If you don't have time for a quick look in the next day or two, please LMK.

jameshiebert:
Cool, thanks for taking a look.

PR #60, opened 2020-02-03
Add Dockerfile

nikola-rados:
The `Dockerfile` is based on `ubuntu:18.04` and installs the necessary `apt` packages, version of python from source and creates a custom user.  The image created by this file is used by the `crmprtd` pipeline to run tests and publish to `pypi`.

**Changes**
- adds a custom `Dockerfile`
- adds `Jenkins` directory for any Jenkins related files


jameshiebert:
I'm a little confused about the purpose of this. It seems that the only thing that the Dockerfile does is to compile Python from source and ensure that it's available. So if someone was to install and run the `crmprtd` Docker file, it wouldn't actually do anything that has to do with `crmprtd`.

Can you clarify why this Dockerfile is necessary for running the Jenkin's pipeline and publishing to PCIC's PyPI? I'm not saying that it's not, but just that it's not evident to me as to why it is from the content of the PR.

nikola-rados:
The image created by the `Dockerfile` is spun up into a container by Jenkins during the pipeline as an environment to run the test suite in.  The pipeline itself takes care of installing `crmprtd` and running the tests.  The items that make this `Dockerfile` unique to `crmprtd` are some of the `apt` packages that are installed:
```
	RUN apt-get update -y && \
	apt-get install -y \
		postgresql-10-postgis-2.4 \ # this
		postgresql-server-dev-10\ # and this
```
`modelmeta` and `pycds` use a similar files.  I originally wanted to place these files in the [`dockerfiles repo`](https://github.com/pacificclimate/dockerfiles) but after some discussion thought it would be better to place them with their respective tools.

jameshiebert:
OK, if the Dockerfile is only used for Jenkins, then I'd lean toward either stashing it away in a `jenkins/` directory (assuming Jenkins can be configured to use an arbitrary Dockerfile) so that's clear that it's not really for *running* `crmprtd`. Or at least put a bright big disclaimer at the top of the Dockerfile that explains what it *is* for. Because generally if there's a Dockerfile in a project, it's for running the software of that project.

With those changes, if it works, feel free to merge away! :)

nikola-rados:
I have the `Dockerfile` and `Jenkinsfile` in a `Jenkins/` directory (I'll change it to lower case).  I can add a comment to the `Dockerfile` to ensure that anyone reading it understands what it is for. :+1:

jameshiebert:
> I have the Dockerfile and Jenkinsfile in a Jenkins/

Bah, OK sorry. I'm complaining about nothing then! This is great, carry on :blush:

PR #69, opened 2020-05-19
Add check for empty observation series for MOTI normalization

eyvorchuk:
This PR checks if an `<observation-series>` is empty and logs a warning if so. An empty series is also included as a test case.


.pytest_cache/v/cache/lastfailed
```diff
@@ -0,0 +1 @@
+{}
```
nikola-rados, line 1: Generally we don't want to be adding files from `.pytest_cache/`.  To avoid this we could add `.pytest_cache/*` to the `.gitinore` or just be careful when using `git add`.

.pytest_cache/v/cache/nodeids
```diff
@@ -0,0 +1,10 @@
+[
+  "tests/test_moti_normalize.py::test_normalize_good_data",
+  "tests/test_moti_normalize.py::test_normalize_missing_stn_indexerror",
+  "tests/test_moti_normalize.py::test_normalize_missing_time",
+  "tests/test_moti_normalize.py::test_normalize_bad_time",
+  "tests/test_moti_normalize.py::test_normalize_missing_var_name",
+  "tests/test_moti_normalize.py::test_normalize_missing_value",
+  "tests/test_moti_normalize.py::test_normalize_bad_value",
+  "tests/test_moti_normalize.py::test_normalize_empty_data"
+]
```
nikola-rados, line 10: Same comment as above.

crmprtd/moti/normalize.py
```diff
@@ -28,6 +28,9 @@ def normalize(iterable):
     et = transform(et)
 
     obs_series = et.xpath("//observation-series")
+    if len(obs_series) == 0:
+        log.warning("Empty observation series: xpath search "
+                    "'//observation-series' return no results")
```
nikola-rados, line 33: Two things here.

1. The output of `et.xpath("//observation-series")` is a list (or some kind of iterable) which means we can very easily check if it has content directly.  Using `if not obs_series` will return `True` when the iterable has no items (example below). This is simply a styling suggestion so please feel free to take it or leave it.  The code is understandable as is.

```python3
>>> obs_series = []
>>> if not obs_series:
...     print('warning')
... 
warning
```
2. If there are no observations, there is no data to process in the loop below. We would want to exit this method is the condition is triggered.

tests/test_moti_normalize.py
```diff
@@ -278,3 +278,28 @@ def test_normalize_bad_value():
...
+      </production-center>
+    </source>
+  </head>
+  <data>
+    <observation-series>
+    </observation-series>
+  </data>
+</cmml>''' # noqa
+    rows = [row for row in normalize(BytesIO(lines))]
+    assert len(rows) == 0
```
nikola-rados, line 305: Awesome, thank you for adding testing right away!

crmprtd/moti/normalize.py
eyvorchuk: Which exit command should I use?

eyvorchuk:
> I noticed that the CI tests were failing from `flake8` issues.

I noticed those right after I did the push and fixed them. The CI tests should be fine now.

crmprtd/moti/normalize.py
nikola-rados: I think you can just use `return`.  Using an empty `return` is the same as returning `None`.  You may want to test this a bit. I imagine there may be some work to do in `process.py` if `None` is returned rather than a generator.

tests/test_moti_normalize.py
```diff
@@ -278,3 +278,28 @@ def test_normalize_bad_value():
...
+        <sub-center>AWP</sub-center>
+      </production-center>
+    </source>
+  </head>
+  <data>
+    <observation-series>
+    </observation-series>
+  </data>
+</cmml>''' # noqa
+    rows = [row for row in normalize(BytesIO(lines))]
```
jameshiebert, line 304: I'll echo what @nikola-rados said: great to have testing right from the start!

I do, however, think that you need two additional checks here. You want to test that:
1. The warning ("Empty observation series"...) that you *want* to be generated, actually is.
1. The *error* ("Could not detect the station id"...) that you *don't* want to be generated, is not.

Look at `pytest`'s warning capture and [`pytest.raises`](https://docs.pytest.org/en/latest/reference.html?highlight=pytest.raises#pytest.raises) for each respectively.

nikola-rados: Ah, good call @jameshiebert!

jameshiebert: I misspoke... I meant [`pytest.caplog`](https://docs.pytest.org/en/latest/reference.html?highlight=pytest.raises#caplog).

crmprtd/moti/normalize.py
```diff
@@ -26,8 +26,11 @@ def normalize(iterable):
     file_stream = iterable_to_stream(iterable)
     et = xmlparse(file_stream)
     et = transform(et)
-
     obs_series = et.xpath("//observation-series")
+    if not len(obs_series[0]):
```
jameshiebert, line 30: I think that your test here doesn't account for the fact that there could be (and I *think* there often are) multiple `observation-series` elements in a data stream. So if the first series is empty, your test will skip all subsequent series, whether or not they are empty. I'll push a change that demonstrates this in the tests.

.gitignore
```diff
@@ -48,6 +48,7 @@ htmlcov/
 nosetests.xml
 coverage.xml
 *,cover
+.pytest_cache/
```
nikola-rados, line 51: :+1:

crmprtd/moti/normalize.py
```diff
@@ -26,9 +26,12 @@ def normalize(iterable):
     file_stream = iterable_to_stream(iterable)
     et = xmlparse(file_stream)
     et = transform(et)
-
     obs_series = et.xpath("//observation-series")
     for series in obs_series:
+        if not len(series):
+            log.warning("Empty observation series: xpath search "
+                        "'//observation-series' return no results")
+            continue
```
nikola-rados, line 34: This should do it, nice work!

jameshiebert:
Bingo! looks great!

PR #70, opened 2020-05-27
Add Github Actions

nikola-rados:
Adds github actions to the repo.  The workflows will take over the responsibilities of Jenkins and Travis which are:
- run python test suite (on all pushes)
- build and publish pypi package (on release creation)

I do not have permission to access the webhooks in this repo but Travis and Jenkins need to be deactivated.

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -5,7 +5,7 @@ on: push
 jobs:
   test:
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
     strategy:
       matrix:
         python-version: [3.6, 3.7]
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -18,6 +18,7 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
       run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-tests.yml
```diff
@@ -0,0 +1,31 @@
+name: CI Tests
+
+on: push
+
+jobs:
+  test:
+
+    runs-on: ubuntu-latest
```
jameshiebert, line 8: Do they have the tags for the exact Ubuntu release? I'd assume that "latest" moves. *This* repo isn't sensitive to system requirements (e.g. GDAL), but others might be.

.github/workflows/ci-tests.yml
```diff
@@ -0,0 +1,31 @@
...
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        sudo apt-get install postgresql-10-postgis-2.4
+        pip install -U pip pytest
+        pip install -r requirements.txt -r test_requirements.txt
+        pip install .
+      env:
```
jameshiebert, line 25: Super minor, but can the `env` section be above the `run` section? Might help it read better since the `run` commands would utilize the `env`.

PR #81, opened 2020-09-16
Feature/bch download

### LINKED ISSUE #78: Write a `download` module for BC Hydro
Our BC Hydro feed is currently being downloaded with some one-off bash scripts. It's not super complicated (basically just a bunch of files off of an FTP site), so we should be able to easily recreate this within the `crmprtd` ecosystem.

See the [`crmprtd.wmb.download`](https://github.com/pacificclimate/crmprtd/blob/master/crmprtd/wmb/download.py) module for a similar approach.

cairosanders:
closes #78
note: using pysftp 0.2.8 instead of latest version. Connection doesn't seem to work with key for 0.2.9 based on what I've read in forums.


cairosanders:
I'm struggling to solve the last few white space/ formatting errors I'm getting from the tests...

crmprtd/bc_hydro/download.py
```diff
@@ -0,0 +1,70 @@
...
+
+            with SpooledTemporaryFile(
+                    max_size=int(os.environ.get('CRMPRTD_MAX_CACHE', 2**20)),
+                    mode='r+') as tempfile:
+
+                for filename in os.listdir(os.getcwd()):
+                    name, extension = os.path.splitext(filename)
+                    if extension == '.zip':
+                        zip_file = ZipFile(filename, 'r')
+                        zip_file.extractall()
```
jameshiebert, line 33: You don't actually need to extract the files from the zip file and write them to disk. The `Zipfile` class provides a `read()` method which is a read interface that does the extraction in memory. See [the examples](https://docs.python.org/3/library/zipfile.html#zipfile.ZipFile.open) for how to use it. Then you can basically dispense with `SpooledTemporaryFile` above.

crmprtd/bc_hydro/download.py
```diff
@@ -0,0 +1,70 @@
...
+                        for txt_file in os.listdir(os.getcwd()):
+                            name, extension = os.path.splitext(txt_file)
+                            if extension == '.txt':
+                                file = open(txt_file, mode='r')
+                                tempfile.write(file.read())
```
jameshiebert, line 38: Again, you don't need to write to the `tempfile`; you can write straight to stdout.

crmprtd/bc_hydro/download.py
```diff
@@ -0,0 +1,70 @@
...
+                                os.remove(os.getcwd() + '/' + txt_file)                
+                tempfile.seek(0)
+                for line in tempfile.readlines():
+                    sys.stdout.buffer.write(line.encode('utf-8'))
+
+    except Exception as e:
```
jameshiebert, line 44: This is good to have a catchall at the bottom, but OTOH it's pretty vague. What kind of exceptions would you expect? What kind of exceptions are possible?

Good use of the `log.exception`. That will ensure that the entire traceback gets logged so at least the user has specific information.

crmprtd/bc_hydro/download.py
```diff
@@ -0,0 +1,70 @@
+# Downloads data from BC hyrdo
+
+import pysftp
```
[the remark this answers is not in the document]
jameshiebert: Looks like you got the `requirements.txt` covered in the mean time, but you still need setup.py.

jameshiebert (commented):
I know that you still have this in "draft" mode, but I wanted to give you some early feedback if you'd like.

Overall, it's looking great! Good job getting the `pysftp` library working, good use of temporary directories (that get cleaned up automatically) and you definitely captured the essence of what we want to do! Way to go.

There are a few things that I would change:
1. hitting disk when you don't need to
1. the auth_key argument conflicts with usage elsewhere in the `crmprtd` suite
1. we don't necessarily want to download *everything* that's available
1. there's no testing yet, but that will require more mocking (to mock out the FTP service), so don't try to take that on.

But with a little more work, we should be able to get it :100: % of the way :smile:

crmprtd/bc_hydro/download.py
```diff
@@ -0,0 +1,70 @@
...
+log = logging.getLogger(__name__)
+
+
+def download(username, auth_key, ftp_server, ftp_dir):
+    # Connect FTP server and retrieve file
+    try:
+        sftp = pysftp.Connection(ftp_server, username=username, 
+                                 private_key = auth_key)
+        with TemporaryDirectory() as tmp_dir:
+            sftp.get_r(ftp_dir, tmp_dir)
```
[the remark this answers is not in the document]
cairosanders: what range should I use?

crmprtd/bc_hydro/download.py
cairosanders: I just did what was in the other download modules for this exception

crmprtd/bc_hydro/download.py
```diff
@@ -0,0 +1,70 @@
...
+                tempfile.seek(0)
+                for line in tempfile.readlines():
+                    sys.stdout.buffer.write(line.encode('utf-8'))
+
+    except Exception as e:
+        log.exception("Unable to process ftp")
+
+
+def main():
+    desc = globals()['__doc__']
```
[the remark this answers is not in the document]
cairosanders: Just want to make sure I'm clear on what a docstring is.Is it the comment at the beginning such as the one in wamr.download? :
```python
'''Downloads the BC Weather and Air Monitoring and Reporting (WAMR) data.

The BC Ministry of Environment, Air Quality Network (AQN) (also known
as WAMR) branch posts a rolling window (one month) of weather data
once every day.

It is recommended to run this script once per one or two weeks to
collect all available data (plus, presumably some duplicate data from
the last run). If the script is run less than once per month, you will
miss data.
'''
```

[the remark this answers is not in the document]
jameshiebert: Bingo. [Read this](https://www.python.org/dev/peps/pep-0257/) for more background. Docstrings can document a module (as is the case here), a class, a function, or even a variable.

crmprtd/bc_hydro/download.py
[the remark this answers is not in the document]
jameshiebert: Make it a parameter that the user can supply from the command line.

crmprtd/bc_hydro/download.py
```diff
@@ -0,0 +1,116 @@
+''' Downloads data from BC hyrdo
+
+BC Hydro posts a rolling window (3 months) observing hourly data
+once a week. Data is in txt files.
+
+It is recommended to run this script once every 3 months to collect
```
jameshiebert, line 6: Well, if they post once a week, it would be better to run once a week to get updates. Then there's a very loose connection between the update cycle and the *you-will-lose-data* cycle. If we recommend to run once a week, then monitoring can be much more casual.

crmprtd/bc_hydro/download.py
```diff
@@ -0,0 +1,116 @@
...
+from tempfile import TemporaryDirectory
+import logging
+import os
+import sys
+import re
+from zipfile import ZipFile
+from argparse import ArgumentParser
+from datetime import date, timedelta
+
+from crmprtd import logging_args, setup_logging, common_auth_arguments
```
jameshiebert, line 22: Do we actually *use* `common_auth_arguments` for anything anymore? I think that this can be pulled.

crmprtd/bc_hydro/download.py
```diff
@@ -0,0 +1,116 @@
...
+def download(username, gpg_private_key, ftp_server, ftp_dir, start_date,
+             end_date):
+
+    # Connect FTP server and retrieve file
+    try:
+        sftp = pysftp.Connection(ftp_server, username=username,
+                                 private_key=gpg_private_key)
+    except Exception:
+        log.exception("Invalid ftp authentication")
+        sys.exit(1)
```
jameshiebert, line 36: We don't typically `sys.exit(1)` from programs. Raising the `Exception` is sufficient, since it exits the program (with a non-zero error code) *and* prints out the stack trace so that there is some actionable information.

crmprtd/bc_hydro/download.py
```diff
@@ -0,0 +1,116 @@
...
+
+    try:
+        with TemporaryDirectory() as tmp_dir:
+            os.chdir(os.path.expanduser(tmp_dir))
```
jameshiebert, line 40: I'd much, much prefer to *not* `os.chdir`. It's a side-effect that is not necessary. Anywhere that you use `os.getcwd()` below, you could just as well have passed `tmp_dir` in and not have changed the directory for the current user. Try to avoid state as much as possible.

crmprtd/bc_hydro/download.py
```diff
@@ -0,0 +1,116 @@
...
+                name, extension = os.path.splitext(filename)
+                if extension == '.zip':
+                    zip_file = ZipFile(filename, 'r')
+                    for name in zip_file.namelist():
+                        txt_file = zip_file.open(name)
+                        sys.stdout.buffer.write(txt_file.read())
+
+    except IOError:
+        log.exception("Unable to download or open files")
+        sys.exit(1)
```
jameshiebert, line 53: Again, try to avoid `sys.exit`.

crmprtd/bc_hydro/download.py
```diff
@@ -0,0 +1,116 @@
...
+    # Add files within date range to temp dir
+    def matchFile(self, filename):
+        match = re.search(r'[0-9]{6}', filename)
+        if match:
+            date = int(match.group(0))
+        if date >= self.start_date and date <= self.end_date:
+            self.sftp.get(filename)
+
+
+def matchDir(dir):
```
jameshiebert, line 72: If this (and `unknownType`) are currently unimplemented and unused, it's best to leave them out of the commit.

crmprtd/bc_hydro/download.py
```diff
@@ -0,0 +1,116 @@
...
+                    for name in zip_file.namelist():
+                        txt_file = zip_file.open(name)
+                        sys.stdout.buffer.write(txt_file.read())
+
+    except IOError:
+        log.exception("Unable to download or open files")
+        sys.exit(1)
+
+
+class DateRange():
```
jameshiebert, line 56: Not meaning to be snarky, but whenever I see a class definition with two methods, and one of them is `__init__` I can't help but refer back to this [classic PyCon talk](https://www.youtube.com/watch?v=o9pEzgHorH0) which advises against such a pattern (see *at least* 2:20-3:20).

You can probably do the same thing with a single method that has four args:

```python
def matchFile(sftp, filename, start_date, end_date):
    ...
```

and it will be more stateless and significantly more testable.

jameshiebert (changes requested):
A few short comments... it's looking a *lot* better, but I think there a few critical things to change before we merge. Thanks so much for working on this!

crmprtd/bc_hydro/download.py
cairosanders: I understand. I did this because I wasn't sure how to pass the filename and 3 more arguments to matchFile through connection.walktree and this was my work around.

crmprtd/bc_hydro/download.py
cairosanders: These are because connection.walktree expects [4 arguments](https://pysftp.readthedocs.io/en/release_0.2.8/pysftp.html?highlight=walktree#pysftp.Connection.walktree), 3 of which are function calls. I got an error when I did not include them. Is there something I could do instead?

crmprtd/bc_hydro/download.py
cairosanders: I think we need it for username

crmprtd/bc_hydro/download.py
cairosanders: I remember the reason I did this was mostly because of this argument for [`connection.get`:](https://pysftp.readthedocs.io/en/release_0.2.8/pysftp.html#pysftp.Connection.get_d)
```
localpath (str) – the local path and filename to copy, destination. If not specified, file is copied to local current working directory
```
and if you specify you have to specify an existing file 
`[Errno 2] No such file or directory: '/tmp/tmpcarurmrp/pcic/Test/PCIC_BCHhourly_200914.zip'` 
so I think I would have to make temp files for this which I can do if that is preferable?

crmprtd/bc_hydro/download.py
jameshiebert: Ah OK, I see why you did that.You're working around their API which is maybe has some... interesting design choices. Arguably they should have had no-op defaults for those arguments.

So yes, what you did is totally functional and works great. Stylistically, I'd argue for something *slightly* more succinct.

1. We can use a lambda expression instead of a function definition and we can use the same for both arguments. I.e.

```python3
no_op = lambda x: None
sftp.walktree(ftp_dir, range.matchFile, no_op, no_op)
```

2. I'd define `no_op` within the `download()` function, so that it retains local scope and isn't visible outside of the module (or even the function). That's a slightly picky point, but it helps keep the module namespace nice and clean.

crmprtd/bc_hydro/download.py
jameshiebert: Oh OK, that makes sense. I *think* what you want is [`functools.partial`](https://docs.python.org/3/library/functools.html#functools.partial). So define your function `matchFile()` with signature:

```python3
def matchFile(filename, start_date, end_date):
    ...
```

And then before the `walktree` call, "freeze" the arguments with the parameters that you have:

```python3
callback = functools.partial(matchFile, start_date, end_date)
sftp.walktree(ftp_dir, matchFile, no_op, no_op)
```

I *think* that's a little more elegant (and will hopefully work), but I'll let you work out the exact implementation.

jameshiebert (changes requested):
OK, thanks for the conversation. I better understand your motivation for making some of the implementation choices that you did. I've provided a bit more feedback with suggested changes. But I haven't *run* any of the code, so please do let me know if you have trouble getting my suggestions to work.

cairosanders:
I had some trouble with the path for `connection.get` because it seems you have to use an existing file in an existing directory, *but* what I did is working now. After I thought about it, I figured I didn't need to make the files using tempfile.TemporaryFile because they're already in a temporary directory. Additionally, the error in the tests is a [flake8 error for `no_op = lambda x: None`](https://github.com/pacificclimate/crmprtd/pull/81/checks?check_run_id=1146423557#step:6:230) so I just left it.

jameshiebert:
Hey @cairosanders hope that you don't mind me finishing this up for you. I didn't want to distract from your DACCS work, and there were some places (i.e. testing) where I didn't necessarily want you to drop everything to work on.

I've added some changes to this branch that I think cover all of the bases. I'd like for you to look over the changes (and the final outcome) and ensure that you understand them. And if you could run the code as well to ensure that it works for you, that would be super helpful.

jameshiebert:
PS I force pushed the branch after rebasing (to pick up all the work on `master` since you started this), so you'll probably have to delete your local `feature/bch-download` branch before pulling it down from GitHub again.

cairosanders:
@jameshiebert Thanks for doing it. Makes sense to me and it works :+1:

PR #82, opened 2020-09-23
Bugfix/77

### LINKED ISSUE #77: Materialized view functions not in user's search path.
When running the `crmprtd_process` script, it's possible to come across the following error:

```
sqlalchemy.exc.ProgrammingError: (psycopg2.errors.UndefinedFunction) function matview_queue_refresh_row(unknown, integer) does not exist
LINE 1: SELECT matview_queue_refresh_row('collapsed_vars_mv', NEW.hi...
               ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
QUERY:  SELECT matview_queue_refresh_row('collapsed_vars_mv', NEW.history_id)
CONTEXT:  PL/pgSQL function "collapsed_vars_mv_obs_raw_insert" line 3 at PERFORM
```

This error shows up if you connect to the database with a user that doesn't have the materialized view functions (e.g. `matview_queue_refresh_row`) in its search path.

We should catch this Exception, and provide a more actionable and informative error message for the user.

cairosanders:
This branch resolves #77 
This is definitely still a work in progress, but I'm hoping to get some feedback on what I've done

### DIFF
--- a/crmprtd/process.py
+++ b/crmprtd/process.py
@@ -73,7 +73,7 @@ def process(connection_string, sample_size, network, is_diagnostic=False):
 
         results = insert(sesh, observations, sample_size)
 
-    except exc.ProgrammingError as e:
+    except exc.ProgrammingError:
         log.exception("Materialized view functions not in user's search path")
 
     log.info('Data insertion results', extra={
--- a/crmprtd/process.py
+++ b/crmprtd/process.py
@@ -4,6 +4,7 @@ from sqlalchemy import create_engine
 from sqlalchemy.orm import sessionmaker
 import logging
 from argparse import ArgumentParser
+from sqlalchemy import exc
 
 from crmprtd.align import align
 from crmprtd.insert import insert
@@ -61,7 +62,8 @@ def process(connection_string, sample_size, network, is_diagnostic=False):
 
     try:
         observations = [
-            ob for ob in [align(sesh, row, is_diagnostic) for row in rows] if ob
+            ob for ob in [align(sesh, row, is_diagnostic)
+                          for row in rows] if ob
         ]
 
         if is_diagnostic:
@@ -71,8 +73,8 @@ def process(connection_string, sample_size, network, is_diagnostic=False):
 
         results = insert(sesh, observations, sample_size)
 
-    except Exception as e:
-        log.exception("Materialized view functions not in search path")
+    except exc.ProgrammingError as e:
+        log.exception("Materialized view functions not in user's search path")
 
     log.info('Data insertion results', extra={
         'results': results, 'network': network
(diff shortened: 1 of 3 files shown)


tests/test_fixture.py
```diff
@@ -13,6 +15,20 @@ def test_db_has_data(test_session):
...
 
+def test_matview_path_exception(test_path_exception):
+    with pytest.raises(exc.ProgrammingError) as e:
+        brandy_hist = History(station_name='Brandy')
+        moti = Network(name='MoTIe')
+        station = Station(native_id='11091', network=moti,
+                          histories=[brandy_hist])
+
+        test_path_exception.query(Station.native_id)
+        test_path_exception.add(station)
```
cairosanders, line 26: I wasn't sure how to implement process here, but tested a query which is what I believe is used in `align()` and insert which is used in `insert()` right now this raises a permission denied rather than `sqlalchemy.exc.ProgrammingError: (psycopg2.errors.UndefinedFunction)`

crmprtd/process.py
```diff
@@ -59,16 +60,22 @@ def process(connection_string, sample_size, network, is_diagnostic=False):
     Session = sessionmaker(engine)
     sesh = Session()
 
-    observations = [
-        ob for ob in [align(sesh, row, is_diagnostic) for row in rows] if ob
-    ]
+    try:
```
cairosanders, line 63: It's still a little unclear to me how to test this manually, so this is a catch all around both align() and insert() right now

tests/conftest.py
```diff
@@ -139,6 +139,14 @@ def test_session(crmp_session, caplog):
     yield crmp_session
 
 
+@pytest.yield_fixture(scope='function')
+def test_path_exception(test_session):
```
jameshiebert, line 143: It's usually more clear to name test fixtures with the name of a noun, since it's a *thing* that you'll be using in your test. E.g. `session_with_user`.

tests/test_fixture.py
jameshiebert: Yeah, so you'll probably want to `assert` something about the text of the Exception that gets raised. E.g. `assert "matview_queue_refresh_row" in e.message` or something like that. That will help you differentiate between the error that we're *trying* to test for, and other errors that shouldn't happen.

jameshiebert: You also want to isolate the `with` block to as narrow a window of code as can possibly raise the error (which should be inserting an observation `Obs`).

jameshiebert (commented):
I have a couple small comments, but stand by because I have breaking news on the issue...

jameshiebert:
At this point, we've overhauled a bunch of things related to this issue/PR including environment variables that set the database schema and migrating the database backend to support materialized views. Closing this PR and issue. It was my mistake, anyway, for marking it as a "good first issue" (it was far too complicated for that).

PR #86, opened 2020-10-08
Update URL of weatheroffice.

### LINKED ISSUE #85: Update Meterorological Service of Canada URL
From ECCC Datamart:

> ======================
> 
> Dear MSC Datamart users,
> 
> Please note that as of October 13, 2020, the following MSC Datamart access URLs will be decommissioned:
> 
>   *   http://dd.weatheroffice.ec.gc.ca
>   *   http://dd.meteo.ec.gc.ca
> These URLs are obsolete and only the following addresses will now be used to access the MSC Datamart data (*) :
> 
>   *   https://dd.weather.gc.ca/
>   *   https://dd.meteo.gc.ca

The URL [is hardcoded here](https://github.com/pacificclimate/crmprtd/blob/master/crmprtd/ec/__init__.py#L39) and needs to be either parameterized or at least changed.

bhawesh96:
Closes #85


crmprtd/ec/__init__.py
```diff
@@ -36,7 +36,7 @@ def makeurl(freq='daily', province='BC', language='e', time=None):
     freq = 'yesterday' if freq == 'daily' else freq
     fname = '{}_{}_{}_{}.xml'.format(
         freq, province.lower(), time.strftime(fmt), language)
-    str = 'http://dd.weatheroffice.ec.gc.ca/observations/xml/'
+    str = 'https://dd.weather.gc.ca/observations/xml/'
```
BioGeek, line 39: User defined names (the variable `str`) should not  shadow python built-in functions ([str](https://docs.python.org/3/library/stdtypes.html?highlight=str#str)).

jameshiebert:
Thanks for the contribution, but we went with a parameterized implementation.

PR #96, opened 2021-05-25
Revisions and tests

jameshiebert:
I made a few changes to hopefully improve the code:

+ Use a dictionary mapping and set lookup rather than a loop over all items
+ Put the mappings in an external file (to keep the code, just code)
+ Added a test case that ensures the conditional gets tripped when appropriate

Feel free to review and let me know if they look good.

### DIFF
--- a/crmprtd/wamr/normalize.py
+++ b/crmprtd/wamr/normalize.py
@@ -3,10 +3,12 @@ import io
 import logging
 import re
 import csv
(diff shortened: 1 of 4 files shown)


faronium (commented):
Good call on moving the station/network relationships out of code. Look forward to seeing it rolled out to production. Let me know how/if I can help.

PR #102, opened 2022-05-18
I53 standardize time arguments

### LINKED ISSUE #53: Implement consistent time arguments each download script
implement consistent time arguments in the download scripts that error out if a time range is selected that is incompatible with its network.

See #50

mariewhibley13:
This branch updates time formats to resolve #53.
Looking for some feedback on my changes.

### DIFF
--- a/crmprtd/moti/download.py
+++ b/crmprtd/moti/download.py
@@ -26,6 +26,8 @@ from warnings import warn
 from datetime import datetime, timedelta
 from argparse import ArgumentParser
 
+import dateutil.parser
+
 # Local
 import crmprtd.download
 from crmprtd import common_auth_arguments, logging_args, setup_logging
@@ -35,7 +37,7 @@ log = logging.getLogger(__name__)
 
 def verify_date(datestring, default, label="start_time"):
     try:
-        return datetime.strptime(datestring, "%Y/%m/%d %H:%M:%S")
+        return dateutil.parser.parse(datestring)
     except (ValueError, TypeError):
         warn(
             "Parameter {} '{}' is undefined or unparseable. Using the "
@@ -110,8 +112,7 @@ def main():  # pragma: no cover
         "--start_time",
         help=(
             "Alternate time to use for downloading "
-            "(interpreted with "
-            "strptime(format='Y/m/d H:M:S'). "
+            "(interpreted with dateutil.parser.parse)."
             "Defaults to one hour prior to now"
         ),
     )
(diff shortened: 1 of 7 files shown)


crmprtd/moti/download.py
```diff
@@ -35,7 +37,7 @@
 
 def verify_date(datestring, default, label="start_time"):
```
jameshiebert, line 38: Since we're now using an external library to verify that user inputted dates are parsable and interpretable into a datetime object, I'd recommend that we *remove* the `verify_date()` function.

That said, `dateutil.parser`'s errors are not exactly user-friendly to the users of this application. E.g.:

```python
>>> from dateutil.parser import parse
>>> parse("Jan 10, 2000")
datetime.datetime(2000, 1, 10, 0, 0)
>>> parse("11/11/11")
datetime.datetime(2011, 11, 11, 0, 0)
>>> parse("not a date")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/james/.local/share/virtualenvs/crmprtd-Sg6rF-Cq/lib/python3.8/site-packages/dateutil/parser/_parser.py", line 1358, in parse
    return DEFAULTPARSER.parse(timestr, **kwargs)
  File "/home/james/.local/share/virtualenvs/crmprtd-Sg6rF-Cq/lib/python3.8/site-packages/dateutil/parser/_parser.py", line 649, in parse
    raise ValueError("Unknown string format:", timestr)
ValueError: ('Unknown string format:', 'not a date')
```
So maybe instead of removing it altogether, we could move it out of `moti.download`, to `crmprtd.download` and then use it (consistently) across *all* of the download modules? Then each module that needs time parsing could use our internal function which either parses it and returns the parsed result, or gives a user-friendly warning and sets the default.

jameshiebert (changes requested):
Everything looks good, however, see my one comment. This will expand the scope of the issue a bit, but I think that it will improve the code and make it easier to use across the board.

jameshiebert (changes requested):
One small tweak: I'd suggest using Python's [f strings](https://docs.python.org/3/reference/lexical_analysis.html#f-strings) instead of `str.format()` in your changes. After that, I think that we should be good to merge.

mariewhibley13:
> One small tweak: I'd suggest using Python's [f strings](https://docs.python.org/3/reference/lexical_analysis.html#f-strings) instead of `str.format()` in your changes. After that, I think that we should be good to merge.

I switched the formatting to use f strings instead. I didn't add the verify_date() function to crd.download because it has its own verify_dates() function but that function has a different functionality. Do you suggest I add it anyway or leave that download script as it is?

jameshiebert (approved):
Looks great! I'm going to go ahead and merge.

PR #104, opened 2022-05-30
I54 implement time range bkp

### LINKED ISSUE #54: Implement time range selection in process script
implement a time range selection to the process script that could filter observations that our outside of the range. This would cut down on the number of skips that we receive on the networks that post rolling windows (and drastically speed up our insertion phase).

See #50

mariewhibley13:
This branch implements a time range selection to the process script that would resolve #54. 

I'm not sure if there should be default values for the start and end date variables. If there should be default values what should they be? I'm thinking there shouldn't be default values and the start and end dates should be empty if nothing is entered for them. Suggestions?


crmprtd/process.py
```diff
@@ -68,9 +87,19 @@ def process(connection_string, sample_size, network, is_diagnostic=False):
     Session = sessionmaker(engine)
     sesh = Session()
 
-    observations = [
-        ob for ob in [align(sesh, row, is_diagnostic) for row in rows] if ob
-    ]
+    # If a start and end date were given then filter observations
+    # Or if just a start date was given. This won't work
+    # Because start and end date have defaults so the wont be empty
+    if (start_date and end_date) or start_date:
```
jameshiebert, line 93: One way that you could keep your logic conceptually simpler is to use what is essentially negative and positive infinity as your defaults. Then, barring no input from the user (or partial input, like a start date but not an end date), the comparison still works and you don't need to mess around with any conditionals/branches.

crmprtd/process.py
```diff
@@ -132,7 +161,23 @@ def main():
         args.log_conf, args.log_filename, args.error_email, args.log_level, "crmprtd"
     )
 
-    process(args.connection_string, args.sample_size, args.network, args.diag)
+    # Right now start date defaults to one month earlier and
+    # end date defaults to now, but not sure if there
```
jameshiebert, line 165: I'd agree that the default behaviour should be *no* date filtering... i.e. if given no input, the process script should accept and process all of the data that it is given.

jameshiebert (commented):
Just a couple points of clarification on the desired behaviour:

crmprtd/process.py
jameshiebert: See [`datetime.datetime.min` and `max`](https://docs.python.org/3.8/library/datetime.html#datetime.datetime.min).

jameshiebert (changes requested):
The code looks perfect to me. It's a simple and elegant implementation that should drastically improve the performance of some of our networks.

Unfortunately, our test suite doesn't have any tests for the `process()` function, so we can't automatically test whether it works or not. Could you write a couple tests? To be clear, it's not a problem that you created, but if you solve it, it will help your code and the rest of the code base.

The tricky part will be that the input to `process()` comes from `sys.stdin`. You'll have to mock that out, using some strange techniques described [here](https://docs.pytest.org/en/7.1.x/how-to/monkeypatch.html) and [here](https://code-maven.com/slides/python/pytest-mocking-stdin-monkeypatch).

I'd propose that you use one set of input (you can use some input that's defined elsewhere in the test suite), and parametrize the tests across four cases: start_date is null or set and end_date is null or set.

tests/test_process.py
```diff
@@ -0,0 +1,56 @@
...
+
+    script_dir = os.path.dirname(__file__)
+    forestry_data = open(
+        os.path.join(script_dir, "..", "crmprtd", "data", "forestry_data")
+    )
+
+    monkeypatch.setattr("sys.stdin", forestry_data)
+
+    process(
+        "postgresql://monsoon.pcic.uvic.ca:5432/crmp",
```
jameshiebert, line 42: I'll do a more complete review, but one quick change to get your tests passing:

One can't/shouldn't use any external resources in the test suite. Instead of hitting a real live production database, use the [test fixture `crmp_session` (or one of the others)](https://github.com/pacificclimate/crmprtd/blob/master/tests/conftest.py#L49) and access its connection string with `crmp_session.get_bind().url`.

jameshiebert:
Looks great! There were just a couple things to tidy up, but you can see what I did in the last commit that I pushed to the branch. This should be good to go.

PR #105, opened 2022-06-17
Feature/new networks

cdmballantyne:
Added three new networks (yt_firewx, yt_avalanche, and dfo_ccg_lighthouse) as well as updated the datetime matching to work with the dfo_ccg_lighthouse format.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -38,6 +38,7 @@ python-json-logger = "==0.1.9"
 "testing.postgresql" = "==1.3.0"
 requests-mock = "==1.7"
 pytest-mock = "==1.11.1"
+crmprtd = {path = "."}
 
 [packages.requests]
 extras = [ "security",]
--- a/Pipfile
+++ b/Pipfile
@@ -38,7 +38,6 @@ python-json-logger = "==0.1.9"
 "testing.postgresql" = "==1.3.0"
 requests-mock = "==1.7"
 pytest-mock = "==1.11.1"
-crmprtd = {path = "."}
 
 [packages.requests]
 extras = [ "security",]
--- a/crmprtd/ec_swob/download.py
+++ b/crmprtd/ec_swob/download.py
@@ -41,14 +41,6 @@ def get_url_list(
     queue = [base_url]
     sesh = requests.Session()
 
-    # Match something like this: 2019-10-18-1600-bc-env-asw-1a02p-AUTO-swob.xml
-    # or this from DFO: 20220525T0230Z_DFO-CCG_SWOB_1060901.xml
-    datep_standard = date.strftime("%Y-%m-%d-%H")
-    datep_dfo = date.strftime("%Y%m%dT%H30Z")
-    search_pattern = re.compile(
-        rf"({datep_standard}|{datep_dfo}).*swob.*\.xml", re.IGNORECASE
-    )
-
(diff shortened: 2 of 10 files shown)


crmprtd/ec_swob/download.py
```diff
@@ -43,6 +43,7 @@ def get_url_list(
 
     # Match something like this: 2019-10-18-1600-bc-env-asw-1a02p-AUTO-swob.xml
     search_pattern = re.compile(r"{}.*swob\.xml".format(date.strftime("%Y-%m-%d-%H")))
+    search_pattern_dfo = re.compile(rf"{date.strftime('%Y%m%dT%H30Z')}.*SWOB.*\.xml")
```
jameshiebert, line 46: OK, so *here* is the spot where you actually *do* want to ignore case, if you were to incorporate these two patterns into one and match either "SWOB" or "swob". It might be a bit tricky, but I'd propose something like this:

```python3
datep_standard = datetime.strftime("%Y-%m-%d-%H")
datep_dfo = datetime.strftime("%Y%m%dT%H30Z")
search_pattern = re.compile(rf"({datep_standard}|{datep_dfo}).*swob.*\.xml", re.IGNORECASE)
```

The `(pattern_a|pattern_b)` *should* match either pattern. See the `|` special character in the [Regular Expression Syntax docs](https://docs.python.org/3/library/re.html#regular-expression-syntax).

crmprtd/ec_swob/download.py
```diff
@@ -69,7 +70,7 @@ def get_url_list(
 
         # either return them or add to the queue
         for url in urls:
-            if search_pattern.search(url):
+            if search_pattern_dfo.search(url) or search_pattern.search(url):
```
jameshiebert, line 73: If you're able to get the combined pattern working, then this change should be able to be reverted.

crmprtd/ec_swob/download.py
jameshiebert: Given the trickiness here, it would probably be worth breaking this URL matching out into its own function so that we can test the pattern matching independently.

jameshiebert (changes requested):
Hey Charlie,
Most of the changes here are pretty straightforward (adding the YT/NWT networks) and look great. The DFO file matching looks like it still needs a bit of work. The changes you made [broke two existing tests](https://github.com/pacificclimate/crmprtd/runs/6835356444?check_suite_focus=true#step:9:173) which indicates that they will break some existing functionality. So It definitely cannot be merged as is. I've try to provide some constructive comments as to how to fix this. Hope that helps.

Pipfile
```diff
@@ -26,6 +26,7 @@ python-json-logger = "==0.1.9"
 "testing.postgresql" = "==1.3.0"
 requests-mock = "==1.7"
 pytest-mock = "==1.11.1"
+crmprtd = {path = "."}
```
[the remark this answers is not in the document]
jameshiebert: I've fixed this.

crmprtd/ec_swob/download.py
jameshiebert: I've moved the pattern matching out and added a few minimal test cases.

jameshiebert:
This looks good to go! Thanks for your hard work on it!

PR #112, opened 2022-11-24
Improve installation instructions

### LINKED ISSUE #111: Make installation instructions more consistent
README lists 3 installation methods, in the following order:

- Makefile, which uses Pipenv for a dev installation
  - Installs project with `pipenv install -e .`, which changes the Pipfile.lock in a way it should not be.
  - Contains target for installing now irrelevant (and very tricky) old versions of postgresql and postgis.
- Using pip only, for production
- Clone and use Pipenv for a dev installation
  - Installs project with `pipenv install --dev .`, which changes the Pipfile.lock in a way it should not be.

Additionally:
- GitHub workflow python-ci performs Pipenv dev installation
  - Installs project with  `pipenv run pip install .`
  - Installs up-to-date versions of postgresql and postgis

Suggestions:
1. Remove Makefile, which usually turns out to be inflexible and therefore not often used.
2. Change clone-and-install instructions to install with `pipenv run pip install .`.
3. Retain prod installation instructions as-is.
4. Add system-level dependency info (postgres, postgis, libpq-dev), which is only implied in python-ci script.
5. Consider updating python-ci script to use postgres-14 etc? This would be consistent with what we have on workstations.

rod-glover:
Resolves #111


rod-glover:
@jameshiebert , any comment on the TODO list above? One possible item left to do.

rod-glover:
Cancelled latest run of tests because we are only changing README.

PR #116, opened 2022-11-25
Upgrade PyCDS to ver 4.0.0

### LINKED ISSUE #114: Upgrade PyCDS to ver 4.0.0
Metnorth and CRMP will both shortly be migrated to the latest database schema revision. Tests show that the current release of crmprtd is not usable with that revision and must have PyCDS upgraded to ver 4.0.0.

rod-glover:
Resolves #114

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -12,7 +12,7 @@ name = "pypipacificclimate"
 lxml = "==4.6.3"
 Pint = "==0.9"
 psycopg2 = "==2.8"
-pycds = { version = "==3.3.0", index = "pypipacificclimate" }
+pycds = { version = "==4.0.0", index = "pypipacificclimate" }
 python-dateutil = "==2.8"
 pytz = "==2019.3"
 PyYAML = "==5.4"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "cbde138e355cb8da2dec8808c206d4b572b4f919ef720a926d4e74b537a25397"
+            "sha256": "b9d0ee146b34969d3151e95d453fcb835d40d48a4d5d6c081b7b3a4bf7af4653"
         },
         "pipfile-spec": 6,
         "requires": {
@@ -289,11 +289,11 @@
         },
         "importlib-metadata": {
             "hashes": [
-                "sha256:da31db32b304314d044d3c12c79bd59e307889b287ad12ff387b3500835fc2ab",
-                "sha256:ddb0e35065e8938f867ed4928d0ae5bf2a53b7773871bfe6bcc7e4fcdc7dea43"
+                "sha256:d5059f9f1e8e41f80e9c56c2ee58811450c31984dfa625329ffd7c0dad88a73b",
+                "sha256:d84d17e21670ec07990e1044a99efe8d615d860fd176fc29ef5c306068fda313"
             ],
             "markers": "python_version < '3.9'",
-            "version": "==5.0.0"
+            "version": "==5.1.0"
         },
         "importlib-resources": {
             "hashes": [
@@ -476,11 +476,11 @@
         },
         "pycds": {
             "hashes": [
-                "sha256:347b218cb81443ca0b87e71fb3cb481688b13133375d6be3bef8697397268ac3",
-                "sha256:66b7e0f4048919516b0a8d2980e8687056af7debb704bf21e210aa95a402f26d"
+                "sha256:1d6eca5c7e3994b36f3e62ac456bdfd3c6ada2bd8c50be2c5cf107dd85261781",
+                "sha256:9d74ddc9b532c131ac85618d401610849cc72ccbbdec5650ad80b132219d4365"
             ],
             "index": "pypipacificclimate",
-            "version": "==3.3.0"
+            "version": "==4.0.0"
         },
         "pycparser": {
             "hashes": [
@@ -635,6 +635,12 @@
             ],
             "index": "pypi",
             "version": "==1.4.44"
+        },
+        "sqlalchemy-citext": {
+            "hashes": [
+                "sha256:a1740e693a9a334e7c8f60ae731083fe75ce6c1605bb9ca6644a6f1f63b15b77"
+            ],
+            "version": "==1.8.0"
         },
         "tomli": {
             "hashes": [
--- a/crmprtd/align.py
+++ b/crmprtd/align.py
@@ -34,17 +34,25 @@ for def_ in (
 
 def closest_stns_within_threshold(sesh, network_name, lon, lat, threshold):
     query_txt = """
-        WITH stns_in_thresh AS (
-            SELECT history_id, station_id, lat, lon, Geography(ST_Transform(the_geom,4326)) as p_existing,
-                Geography(ST_SetSRID(ST_MakePoint(:x, :y),4326)) as p_new
+        WITH hxs_in_thresh AS (
+            SELECT 
+                history_id, 
+                station_id, 
+                lat, 
+                lon, 
+                Geography(ST_Transform(the_geom, 4326)) as p_existing,
+                Geography(ST_SetSRID(ST_MakePoint(:x, :y), 4326)) as p_new
             FROM crmp.meta_history
-            WHERE the_geom && ST_Buffer(Geography(ST_SetSRID(ST_MakePoint(:x, :y), 4326)),:thresh)
+            WHERE the_geom && ST_Buffer(
+                Geography(ST_SetSRID(ST_MakePoint(:x, :y), 4326)), :thresh
+            )
         )
-        SELECT history_id, ST_Distance(p_existing,p_new) as dist
-        FROM stns_in_thresh
-        NATURAL JOIN crmp.meta_station
-        NATURAL JOIN crmp.meta_network
-        WHERE network_name = :network_name
+        SELECT 
+            history_id, ST_Distance(p_existing, p_new) as dist
+        FROM hxs_in_thresh hx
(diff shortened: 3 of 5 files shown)


rod-glover:
I fixed a few obvious problems that were causing tests to fail. 

The 6 remaining failures all follow this pattern:

```
E   (psycopg2.errors.NotNullViolation) null value in column "standard_name" violates not-null constraint
E   DETAIL:  Failing row contains (1, test_var_b, mm, null, null, null, null, null, null, 1).
E   
E   [SQL: INSERT INTO crmp.meta_vars (net_var_name, unit, standard_name, cell_method, precision, long_description, display_name, short_name, network_id) VALUES (%(net_var_name)s, %(unit)s, %(standard_name)s, %(cell_method)s, %(precision)s, %(long_description)s, %(display_name)s, %(short_name)s, %(network_id)s) RETURNING crmp.meta_vars.vars_id]
E   [parameters: {'net_var_name': 'test_var_b', 'unit': 'mm', 'standard_name': None, 'cell_method': None, 'precision': None, 'long_description': None, 'display_name': None, 'short_name': None, 'network_id': 1}]
```

I am fairly the problem lies in [`infer.create_variable`](https://github.com/pacificclimate/crmprtd/blob/i114-pycds-4.0.0/crmprtd/infer.py#L38-L41), which creates a new variable thus: 
`Variable(network=network, name=variable_name, unit=unit)`

We need to provide values for the newly not-null columns in Variable, specifically `standard_name`,  `cell_method`, `display_name`. I am not sure where to derive this information. It is easy enough add the required arguments, but where the values should come from in real data, which I assume is accurately simulated by test data such as `data/forestry_data.xml`.

@jameshiebert , can you advise on this?

rod-glover:
We could put in dummy values, but that poses the question of why these columns were made not nullable in the first place. Seems like a last resort.

jameshiebert:
Hrmmm, even though I conceived and wrote the `infer` phase after we worked through [the NOT NULL constraints](https://github.com/pacificclimate/pycds/issues/29), I seem to have forgotten about them.

It is definitely not possible to programmatically determine `cell_method` and `standard_name`. Those require human expertise and sometimes conversations between ourselves and the data network managers.

That said, the way the `infer` phase works is that it makes database modifications and then rolls them back. See the [`module docstring`](https://github.com/pacificclimate/crmprtd/blob/master/crmprtd/infer.py#L9) for the explanation.

Given that, I think it would be sufficient to add dummy values for the `infer`-based inserts, since they're not actually going to end up in the database.

rod-glover:
I added placeholders for those values. 
- I used the value `"undefined"` for `cell_method`. Would you prefer something different?
- I have doubts about using `network_name` to fill in `standard_name`; perhaps it should be `"undefined"` too.

jameshiebert:
I think that those are both fine. If we want the code to be more self documenting, I'd say something like: `standard_name="this_will_never_be_committed", cell_method="time: not_committed"`

rod-glover:
After some discussion, @jameshiebert said:

> Here's a state diagram of what I would say are the various inputs (diagnostic, new stations, new variables) and outcomes (whether a commit happens or an exception is raised).
> 
> diagnostic | new stations | new variables | commit | raises
> --- | --- | --- | --- | ---
> False | True | True | no | yes
> False | False | True | no | yes
> False | True | False | yes | no
> False | False | False | irrelevant | no
> True | True | True | no | no
> True | False | True | no | no
> True | True | False | no | no
> True | False | False | irrelevant | no
> 
> That this seems overly complicated signals that maybe there would be a better way to approach the problem. But I definitely see a use case for taking a set of observations and inserting all of the stations that would need to be created. And/or telling you what variables need to be created. Charlie utilized that heavily for her creation of the Yukon and NWT new networks.

rod-glover:
Exception now raised according to table above. A few items to consider, flagged by TODO comments in code:

1. [Raising exception may prevent other useful behaviour (e.g., logging)](https://github.com/pacificclimate/crmprtd/blob/bbe22153c8746d1de6bcaffb37ba2087464a6a10/crmprtd/infer.py#L115)
2. [I think there is a subtle bug that creates station and history recs in diagnostic mode](https://github.com/pacificclimate/crmprtd/blob/bbe22153c8746d1de6bcaffb37ba2087464a6a10/crmprtd/align.py#L135)
3. [When matching histories spatially, they are not ordered by distance. Why? Various things connected with this are labelled "nearest"](https://github.com/pacificclimate/crmprtd/blob/bbe22153c8746d1de6bcaffb37ba2087464a6a10/crmprtd/align.py#L70)
4. A couple of other minor items.

rod-glover:
@jameshiebert , after further thought, I think that (1) ought to be dealt with. I believe that processing the histories/stations and processing the variables is independent (i.e., neither depends on the other), so we can re-order them so that hxs/stns come first, then vars, which might raise an exception. Although hxs/stns won't create records, it does log some useful info and by raising the exception first we lose that.

rod-glover:
I also believe (2) is the case with high confidence:
- The "close stations" search can return zero stations (no close ones).
- In that case `create_station_and_history_entry` is invoked. It (a) does not receive the `diagnostic` parameter, and (b) ignores it anyway, which means that it _always_ creates record(s). 

In diagnostic mode, this would be incorrect, yes?

rod-glover:
Really wondering what happened with (3). 
- The base query returns `history_id`s with distances from the specified lat/lon, ordered by distance. But then it returns the id's as a set, which is unordered. 
- Various names that use the results of this function include "close/near", e.g., `close_stns`, `find_nearest_history`. But they of course are receiving an unordered list, so are just getting a more or less random one.

Did some rethinking happen around the value of nearness when seeking a history with a spatial match? Code bears traces of that, but maybe it is just a coding error or other misunderstanding. If the former, the code (query) can be simplified. If the latter, then there's some work to be done to restore order (ha ha).

rod-glover:
@jameshiebert , I know you are extremely busy right now. I need to proceed with this PR, so I will makeh what I think are sound fixes for (1) and (2), and will create a separate issue for (3). If you disagree with what I've done later, we can always start new PR(s).

rod-glover:
> Previously, insertion activity was tested by a count of the number of successful bulk insertions in the bisection strategy, and the number of actual records inserted was not counted. A [branch](https://github.com/pacificclimate/crmprtd/tree/update-tests) containing just the updates to logging and testing has been pushed and seems to verify the claim that there was a bug which was missed by the test. **_Alternatively_**, I have introduced a bug here and only a subset of the observations presented should be inserted.

To `test_process.py`, I added a bit of code to count the number of observations in the database before and after invoking `process`. I added this to both the new PR branch and the old-code test branch. 

Results: In the old code, only 2360 of the 2510 observations to be inserted were actually inserted, consistent with the count obtained by analyzing the log outputs. In the new code, all 2510 observations were inserted.

Conclusion: There is no reason why that full set of observations _cannot_ be inserted: They all _were_ inserted by the new code, without error. The old code seems for some reason to skip inserting some of the observations available. I have not looked into the reasons for that.

@jameshiebert , do you have any comment on this? I do believe it was a bug in the old code, but you will know it much better than I do. This is the last thing I need resolved before I feel confident in merging the PR.

rod-glover:
Note re. failed check: (Expletives omitted.) 

This identical GH workflow code was able to run the tests (using Pipenv) successfully 4 days ago. As of today, it appears Pipenv cannot now start, so tests cannot be run.

A brief inspection of the log of the Pipenv installation shows:

4 days ago:
```
Successfully installed certifi-2022.9.24 distlib-0.3.6 filelock-3.8.2 pipenv-2022.3.28 platformdirs-2.5.4 virtualenv-20.17.1 virtualenv-clone-0.5.7
```

Today:
```
Successfully installed certifi-2022.12.7 distlib-0.3.6 filelock-3.8.2 pipenv-2022.3.28 platformdirs-2.6.0 virtualenv-20.17.1 virtualenv-clone-0.5.7
```

A different version of `certifi` was installed. It is not at all clear this is the culprit, but thought I would note it while it is in mind.

rod-glover:
The cause of the error when invoking Pipenv was the step in which the Pipenv environment was cached. This had to be removed altogether, not just the outcome ignored, in order to solve the problem.

rod-glover:
@jameshiebert , I traced the bug to an apparently tiny change I made in how the history record is created. Not sure why I did it, but it is now fixed, which is to say restored to its original form. Ready for a review if you have the time.

tests/test_process.py
```diff
@@ -1,54 +1,117 @@
...
+
     process(
-        crmp_session.get_bind().url,
-        50,
-        "bc_forestry",
-        start_date,
-        end_date,
-        False,
-        True,
+        connection_string=crmp_session.get_bind().url,
```
jameshiebert, line 103: Good choice. That makes the test much more clear.

tests/test_process.py
```diff
@@ -1,54 +1,117 @@
...
-            bool("Successfully inserted observations" in str(record))
-            for record in caplog.records
-        )
-        == num_inserts
-    )
+    num_obs_in_db_after = num_observations_in_db(crmp_session)
+    num_obs_inserted_in_db = num_obs_in_db_after - num_obs_in_db_before
+    log_num_obs_inserted = get_num_obs_inserted(caplog.records)
+
+    assert num_obs_inserted_in_db == expected_num_inserts
```
jameshiebert, line 116: Excellent. That makes for a much more robust test!

jameshiebert (commented):
Thanks for your hard work on this PR Rod! There's a lot of excellent attention to detail contained in this, and aside from completing the goal of upgrading the package to the use of PyCDS 4.0.0, it also makes significant improvements to the code.

I believe that I only had a few small suggestions, but overall this is a great PR and I'll be happy to see it merged.

PR #124, opened 2023-01-04
Bring execution scripts under change contro

### LINKED ISSUE #123: Bring execution scripts under change control
At present we have scripts on the `crmprtd` server that execute downloading and processing for the various networks. They are invoked by cron. As a prelude to updating them, bring them under change control.

Questions and considerations: 
- These scripts will be part of each release from hereon in. 
- They should be installed as part of the package installation. [They can be included as part of the `setup.py` `scripts` list.](https://stackoverflow.com/a/50797895)

rod-glover:
Resolves #123 

This only adds the existing scripts. They are now installed in the installation environment `bin` directory (e.g., `venv/bin/`), like all other scripts (which are typically python scripts but can in fact be anything). 

After installing a release that contains this PR, the existing scripts can be deleted from the directory `~/bin/` and the PATH variable should not include that directory. To use them in the same manner as currently set up, the PATH variable should include the environment `bin/`, but that is automatic when a venv is activated.


rod-glover:
@jameshiebert , it would be good to get your eyes on this. No change to the scripts, just to how they are managed.

rod-glover:
I feel fairly confident now about this. Will merge without review.

PR #126, opened 2023-01-05
Support multiple target databases

### LINKED ISSUE #125: Modify execution scripts to feed databases at different revisions
Our deployment practice now includes Blue-Green transitions, in which independent production environments, including databases, are created and tested before public switchover. This includes feeding data to both the old database and the new. There are at least a couple of ways to accomplish this.

Notes:

1. To feed a database at a given revision, we must use a version of `crmprtd` compatible with it. Therefore in general during a transition period, we will be invoking different versions of `crmprtd` to do the work.

### Option A: Download data twice: once for each target database

What needs to happen:
- All cron jobs must refer explicitly to the version of `crmprtd` required for the target database.
- Create separate cron jobs for each target database. (This is where the double downloads come from.)
- After switchover, remove cron jobs for decommissioned target database.

Further thoughts:
- At present, the target database is hardcoded in the execution scripts. That should be abstracted out as argument(s) to the script, and supplied in the crontab entry.
- Log and cache filepaths need to be different; otherwise outputs from different processes (with different target databases) are interleaved indistinguishably in the files. These paths too should be determined (partly) by a script argument.

### Option B: Download data once; process twice

This option would eliminate duplication but:
- The scripts would become more complicated, or at least need to be split up into download and process scripts.
- Updating the crontab becomes more complicated.

### Current crontab 

For reference (long commented-out sections omitted):

```
# new crontab, post monsoon.pcic -> db.pcic migration, 20220912

SHELL=/bin/bash
PATH=~/env_3.5.1/bin:~/bin:/usr/local/bin:/usr/bin:/bin

@hourly                         hourly_swobml2_newdb.sh
@weekly                         wamr_newdb.sh
40 4 * * *                      wmb_newdb.sh
@hourly                         ec_newdb.sh hourly
@daily                          ec_newdb.sh daily
30 * * * *                      moti_newdb.sh
@daily                          crd_newdb.sh
@daily                          bch_newdb.sh
0 7 1,15,28,29,30,31 * *        echo "bash /home/crmprtd/code/crmp_fanslow/read_NRT_ASP_snow_pillow_data_by_variable.sh" | batch
30 2 27-31,1-4 * *              echo "bash /home/crmprtd/code/crmp_fanslow/ec_csv_runner.sh" | batch 
@daily                          echo "psql -h db.pcic.uvic.ca -U crmprtd -d crmp -p 5433 -f ~/bin/update_matviews_lazy.sql" | batch
@monthly                        echo "manage-views.py -d 'postgresql://crmp@db.pcic.uvic.ca:5433/crmp' refresh all" | batch
```

rod-glover:
Resolves #125 

This PR implements Option A (Download data once per database). Notes:
- Deletes scripts not presently in use.
- Renames remaining scripts to eliminate now-unnecessary suffixes.
- Adds arguments to scripts for tag and target database.
- Adds documentation about this to README.


crmprtd/execution/bch.sh
```diff
@@ -1,9 +1,16 @@
...
+# (Not currently used.)
+db="$2"
+
 export net=bc_hydro
 export NOW=$(date +'%Y-%m-%dT%H:%M:%S')
 
-export log_file=~/${net}/logs/${net}_json.log
-export cache_file=~/${net}/cache/${net}_${NOW}.txt
+export log_file=~/${net}/logs/${tag}_{net}_json.log
+export cache_file=~/${net}/cache/${tag}{net}_${NOW}.txt
```
rod-glover, line 15: Add `_` between tag and net.

rod-glover:
@jameshiebert , can you have a look at this? I'd suggest reading the updated README before anything else.

crmprtd/execution/metnorth.sh
```diff
@@ -1,19 +1,17 @@
 #!/bin/bash
```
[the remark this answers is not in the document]
jameshiebert: Yeah, that's a good idea.

crmprtd/execution/wmb.sh
```diff
@@ -1,9 +1,17 @@
...
+
+# Target database DSN, e.g., postgresql://crmprtd@db.pcic.uvic.ca:5433/crmp
+db="$2"
+
 export net=wmb
 export NOW=$(date +'%Y-%m-%dT%H:%M:%S')
 
-export log_file=~/${net}/logs/mof_json.log
-export cache_file=~/${net}/download/${net}_${NOW}.txt
+# TODO: Is `mof` right??? Why not `${net}`?
```
jameshiebert, line 12: "Legacy". This is Ministry of Forests - Wildfire Management Branch. One of the first networks that we incorporated, before we even formalized `crmprtd`. So at the time, there were no naming conventions and that's what it got named. It stuck.

jameshiebert (approved):
Looks great! I didn't do a painstakingly close read of all of the execution specifics, but I very much support your overall goal here, and this is a great way to simplify all of the invocation scripts.

PR #128, opened 2023-01-10
Replace `download_<network>` scripts with a `download -N <network>` script

### LINKED ISSUE #127: Replace `download_<network>` scripts with a `download -N <network>` script
This will make it possible to eliminate some dynamic content from `setup.py`, which makes the upcoming transition to `pyproject.toml` easier/possible. (There is some other dynamic content that will be the subject of a different issue.)

rod-glover:
Resolves #127 

There is very little net change. A single script, `crmprtd_download`, which takes argument `-N/--network` replaces the individual `download -N <network>` scripts.

 Script `crmprtd_download` also defines the common arguments for all of them (e.g., logging options) and passes in its arg parser to each, which adds its specific arguments (unchanged from before) as a child arg parser. Minor adjustments to the function signatures and parser creation accommodate this.

Execution scripts are likewise adjusted. This was a simple search/replace and they read almost exactly as before.

Other notes:
- Previous `download` module is renamed `download_utils`, because that's what it is and we want `download` for our new thing.
- Added docstrings to the per-network download scripts. Bonus: Added type hinting on these because hey, why not and it's easier than adding it to the docstrings.


tests/test_scripts.py
```diff
@@ -10,9 +17,10 @@
         for name, ep in crmprtd.pkg_resources.get_entry_map("crmprtd")[
             "console_scripts"
         ].items()
+        if name != "crmprtd_download"
```
jameshiebert, line 20: At this point the set of entry points excluding `crmprtd_download` is only of length 2. Maybe simpler to just enumerate them?

jameshiebert (approved):
Looks good to me! This sets us up nicely for the `pyproject.toml` migration.

PR #134, opened 2023-01-14
Replace execution shell scripts with a Python script

### LINKED ISSUE #129: Add dispatcher execution script on model of `crmp_download`
Using a network argument, script `crmp_download` dispatches to the various per-network download scripts. We should do the same with the execution scripts. This will make `setup.py` fully static, setting us up for an easy transition to `pyproject.toml`.

rod-glover:
Resolves #129

This PR:
- Moves all the functionality of the execution scripts into a Python script.
- Extends `infill.download_and_process` to handle all cases of optionally downloading, caching, and processing data. It does not change `infill.infill` or script `infill_all.py`.
- Uses the extended `download_and_process` to implement the execution scripts in Python.

Notes:
- Implementing the execution scripts in Python reveals that there are several inconsistencies across networks in the log and cache filenames. This makes the code needlessly complicated and probably makes it harder to use (i.e., name and access) those actual files elsewhere. I propose these inconsistencies be removed. At the same time, we could rename the existing ones so that all existing and future ones are consistent. Unless of course those files are already in use by those names elsewhere ...


rod-glover:
TODO: Write tests for dispatch actions (actual calls to download_and_process).

rod-glover:
TODO:

- [ ] Remove shell files (execution scripts)

crmprtd/download_cache_process.py
```diff
@@ -0,0 +1,371 @@
...
+        "bc_tran",
+        "dfo_ccg_lighthouse",
+        "nt_forestry",
+        "nt_water",
+        "yt_gov",
+        "yt_water",
+        "yt_avalanche",
+        "yt_firewx",
+    ):
+        return f"~/{network_name}/cache/{tag}_{network_name}_{ts}.xml"
```
rod-glover, line 114: @jameshiebert , if we were to settle on either `cache/` or `download/` as a directory path component, that would simplify this function considerably. Would there be any adverse impact to doing that? Once we deployed this version, we'd have to rename the directories on server crmprtd. And update `infill_all`.

crmprtd/download_cache_process.py
```diff
@@ -0,0 +1,371 @@
...
+
+    parser.add_argument(
+        "-D",
+        "--describe",
+        nargs=1,
+        choices=NETWORKS + network_alias_names,
+        action=OneAndDoneAction,
+        function=describe_network,
+        help="Describe a network name or alias",
+    )
```
rod-glover, line 323: This option should be pulled out into a separate script, `crmprtd_describe`

tests/test_infill.py
```diff
@@ -0,0 +1,142 @@
...
+    download_and_process,
+    round_datetime,
+    datetime_range,
+    chain_subprocesses,
+)
+
+
+# !! IMPORTANT !!
+#
+# Because crmprtd.infill imports run (and not just subprocesses), we must mock
```
rod-glover, line 19: Fix wording here.

crmprtd/download_cache_process.py
```diff
@@ -0,0 +1,371 @@
...
+        )
+    else:
+        now = datetime.datetime.now()
+        download_and_process(
+            network_name=network_name,
+            log_args=log_args(**kwargs),
+            download_args=download_args(**kwargs, time=now),
+            cache_filename=cache_filename(**kwargs, timestamp=now),
+            connection_string=connection_string,
+        )
```
rod-glover, line 245: Is moti also a special case? `infill.infill` has this:

```python
    if "moti" in networks:

        # Query all existing stations
        stations = get_moti_stations(connection_string)

        # MoTI has an insane config where each user has a specific set
        # of stations associated with it. PCIC wants *all* the
        # stations which is too many for one request, so we have two
        # users with half of the stations. This way of requesting the
        # data will get many "skips", but at least we'll get all of
        # the data.
        for auth_key in ("moti", "moti2"):
            # Divide range into 6 day intervals
            for interval_start, interval_end in zip(
                weekly_ranges[:-1], weekly_ranges[1:]
            ):
                for station in stations:
                     ...
```

but the execution script simply had this:

```bash
echo "crmptd_download -N ${net} -u https://prdoas5.apps.th.gov.bc.ca/saw-data/sawr7110 --auth_fname ~/.rtd_auth.yaml --auth_key=${net} -L ~/logging.yaml --log_filename ${log_file} | tee ${cache_file} |\
 crmprtd_process -c ${db} -N ${net} -L ~/logging.yaml --log_filename ${log_file}" | batch
```

Why are these so different?

rod-glover (commented):
@jameshiebert , some questions for you below.

crmprtd/download_cache_process.py
rod-glover: On second thought, this should stay where it is, as it includes aliases, which are only valid with respect to `download_cache_process`

crmprtd/download_cache_process.py
```diff
@@ -0,0 +1,371 @@
...
+    :param province: Province code (2 letters) to include in filename (network ec only).
+    :param _: Ignored additional kw args.
+    :return: Filename.
+    """
+    check_network_name(network_name)
+    if network_name in ("ec",):
+        return f"~/{network_name}/logs/{tag}_{province.lower()}_{frequency}_json.log"
+    if network_name in ("wmb",):
+        # Oh, yes, this one has to be a special snowflake.
+        return f"~/{network_name}/logs/{tag}_mof_json.log"
```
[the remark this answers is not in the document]
jameshiebert: Sure that's fine. Just symlink on the VM.

crmprtd/download_cache_process.py
```diff
@@ -0,0 +1,371 @@
...
+    :param frequency: Download frequency to include in filename (network ec only).
+    :param province: Province code (2 letters) to include in filename (network ec only).
+    :param _: Ignored additional kw args.
+    :return: Filename.
+    """
+    check_network_name(network_name)
+    if network_name in ("ec",):
+        return f"~/{network_name}/logs/{tag}_{province.lower()}_{frequency}_json.log"
+    if network_name in ("wmb",):
+        # Oh, yes, this one has to be a special snowflake.
```
jameshiebert, line 69: :smirk: Well if they didn't change their ministry name every 6 months...

crmprtd/download_cache_process.py
```diff
@@ -0,0 +1,371 @@
...
+
+network_alias_names = tuple(network_aliases.keys())
+
+
+def check_network_name(network_name):
+    if network_name not in NETWORKS:
+        raise ValueError(f"Network name '{network_name}' is not recognized.")
+
+
+def log_filename(
```
jameshiebert, line 47: On one hand, I'm hesitant to codify the output filepaths... but OTOH if a user requires flexibility, then they still have the ability to invoke the download scripts directly.

crmprtd/download_cache_process.py
```diff
@@ -0,0 +1,371 @@
...
+    :param network_name: Network name.
+    :param _: Remainder args. Passed through.
+    :return:
+    """
+    check_network_name(network_name)
+
+    common_args = f"-N {network_name} ".split()
+    net_args = None
+
+    # Set net_args per network.
```
jameshiebert, line 160: I think that all of these need to be documented in the online help somewhere. *Especially* the auth arguments. (By necessity,) There are a lot of implied requirements that someone could easily overlook if they went to run this on a fresh system (or, say, a docker image in a K8s cron job).

crmprtd/download_cache_process.py
jameshiebert: Technically *every* network is a "special case". That's why this software exists... to allow us to download data according to the specific ways that each network expose it to us.

In the case of MoTI, there are two endpoints:

1. The default endpoint which gives the last hour of data from all stations
2. A "historic" endpoint where you can provide a station id and time range (no greater than 7 days) and get historic data. The infill script is specifically tailored to this use case for obvious reasons.

jameshiebert: Because of this, running a historic query is much more heavyweight than running regular hourly updates from the default endpoint (and its preferable to use the latter). Can you think of a good way to encode this into the script behavior and/or the invocation options?

crmprtd/infill.py
```diff
@@ -209,17 +321,18 @@ def datetime_range(start, end, resolution="hour"):
...
+    resulted in many HTTP 500 errors, so we're using a conservative "week" of 6 days.
 
-    The loosely defined 'month' resolution returns 28 day intervals
-    (for CRD).
+    The loosely defined 'month' resolution returns 28-day intervals
+    (for use by CRD).
 
     Args:
-        start (datetime.datetime): The beginnng of the range
+        start (datetime.datetime): The beginning of the range
```
jameshiebert, line 335: If only it were easy to use emacs spell check on code files :)

setup.py
```diff
@@ -39,13 +38,10 @@ def run_tests(self):
...
         "scripts/moti_insert_files.py",
         "scripts/moti_infill_insert.py",
-    ]
-    + [
-        f"crmprtd/execution/{name}"
-        for name in pkg_resources.resource_listdir("crmprtd", "execution")
     ],
     entry_points={
         "console_scripts": [
+            "crmprtd_dcp=crmprtd.download_cache_process:main",
```
jameshiebert, line 44: I don't *love* the name `crmprtd_dcp`... is there something that's easier to remember that we can name it?

jameshiebert (commented):
Impressive work Rod. Looks good overall, though I have a few suggestions for improvement here and there.

crmprtd/download_cache_process.py
rod-glover: We can allow the user to specify it with an optional log/cache filepath. If not specified, it can fall back to this.

crmprtd/download_cache_process.py
rod-glover: Point taken re. "special cases". 

The info on MoTI is very helpful, thanks. I think it should be feasible to allow for both historical infill and hourly updates. That said, that's what `infill_all` still does. I didn't change its behaviour. Options:
1. Leave d-c-p and infill_all separate, and use them separately. No effort beyond what's already been done; a bit clunky but it works.
2. Merge infill_all functionality into d-c-p. More effort; possibly a nicer interface.

Preference?

setup.py
rod-glover: Hmm. Tougher than it sounds, without resorting to `download_cache_process`

`crmp_dl_c_proc`? Not exactly easier to remember, and clunkier.

`crmp_pipeline`? "Pipeline" is already taken, to mean various parts of the processing pipeline. But this is quite literally a pipeline too.

`crmp_everything`? Kind of like a "bagel with everything"? (Plus, bonus wink at currently insanely popular movie.)

`crmp_all`? Ditto, except no movie ref.

crmprtd/download_cache_process.py
rod-glover: I've created a separate [issue](https://github.com/pacificclimate/crmprtd/issues/136) for this. The question is more general than just for MoTI, although I think that is the gnarliest one.

crmprtd/download_cache_process.py
rod-glover: Yes. I think a "dry-run" flag, which prints rather than executes the commands, would be useful. It's straightforward. I'll try it and you can see what you think.

rod-glover: @jameshiebert , here's some sample outputs of the dry-run option for `crmprtd_dcp`. I'm intending this to fill in for the rather complicated (and needing ongoing maintenance) documentation we want.

- Some have the processing step and others do not, distinguished by whether the `-c` (connection string) option is specified. 
- I edited out the `-u` argument value in case it is secret. Is it?
- Comments on format? These lines could be fed straight into a Linux command line and they would work. But maybe not so readable?

```
(crmprtd) rglover@pcic-3002:~/code/crmprtd$ crmprtd_dcp --dry_run -N moti -T tagg
crmprtd_download -N moti -u <secret?> --auth_fname ~/.rtd_auth.yaml --auth_key=moti -L ~/logging.yaml --log_filename ~/moti/logs/tagg_moti_json.log > ~/moti/cache/tagg_moti_2023-01-19T16:53:27.txt
(crmprtd) rglover@pcic-3002:~/code/crmprtd$ crmprtd_dcp --dry_run -N ec -F daily -T tag -c DSN
crmprtd_download -N ec -p bc -F daily -L ~/logging.yaml --log_filename ~/ec/logs/tag_bc_daily_json.log | tee ~/ec/cache/tag_daily_bc_2023-01-19T16:54:04.xml | crmprtd_process -N ec -c DSN -L ~/logging.yaml --log_filename ~/ec/logs/tag_bc_daily_json.log
crmprtd_download -N ec -p yt -F daily -L ~/logging.yaml --log_filename ~/ec/logs/tag_yt_daily_json.log | tee ~/ec/cache/tag_daily_yt_2023-01-19T16:54:04.xml | crmprtd_process -N ec -c DSN -L ~/logging.yaml --log_filename ~/ec/logs/tag_yt_daily_json.log
(crmprtd) rglover@pcic-3002:~/code/crmprtd$ crmprtd_dcp --dry_run -N ytnt -T tag
crmprtd_download -N nt_forestry -d "2023/01/19 23:00:00" -L ~/logging.yaml --log_filename ~/nt_forestry/logs/tag_nt_forestry_json.log > ~/nt_forestry/cache/tag_nt_forestry_2023-01-19T15:54:56.xml
crmprtd_download -N nt_water -d "2023/01/19 23:00:00" -L ~/logging.yaml --log_filename ~/nt_water/logs/tag_nt_water_json.log > ~/nt_water/cache/tag_nt_water_2023-01-19T15:54:56.xml
crmprtd_download -N yt_gov -d "2023/01/19 23:00:00" -L ~/logging.yaml --log_filename ~/yt_gov/logs/tag_yt_gov_json.log > ~/yt_gov/cache/tag_yt_gov_2023-01-19T15:54:56.xml
crmprtd_download -N yt_water -d "2023/01/19 23:00:00" -L ~/logging.yaml --log_filename ~/yt_water/logs/tag_yt_water_json.log > ~/yt_water/cache/tag_yt_water_2023-01-19T15:54:56.xml
crmprtd_download -N yt_avalanche -d "2023/01/19 23:00:00" -L ~/logging.yaml --log_filename ~/yt_avalanche/logs/tag_yt_avalanche_json.log > ~/yt_avalanche/cache/tag_yt_avalanche_2023-01-19T15:54:56.xml
crmprtd_download -N yt_firewx -d "2023/01/19 23:00:00" -L ~/logging.yaml --log_filename ~/yt_firewx/logs/tag_yt_firewx_json.log > ~/yt_firewx/cache/tag_yt_firewx_2023-01-19T15:54:56.xml
```

setup.py
rod-glover: I'm gonna go with `crmprtd_pipeline`. It's easy enough to change later if we don't like it.

crmprtd/download_cache_process.py
rod-glover: I think this works well, and it is self-maintaining.

PR #140, opened 2023-01-28
Convert to pyproject.toml

### LINKED ISSUE #135: Convert to pyproject.toml
This will be an experiment. Even a false start on this was instructive. We now know that the easiest way to do this is to avoid any dynamic content in the `pyproject.toml`, even if there is a provision for it in the spec. That's because it's not clear how to get the build tool to fill in the dynamic content, which is what the spec says should happen.

Recent changes to `crmprtd` have removed the need for dynamic content, so we can go ahead with the simple approach.

rod-glover:
Resolves #135 

This PR:
- Converts project metadata management entirely to `pyproject.toml`. 
- Removes `setup.py`.
- Omits the following CLI scripts (but preserves their code) because they will need a small amount of refactoring to be runnable as entrypoints. It's not clear which if any of these need to become CLI scripts. Happy to refactor as needed:
  - `ec_recovery.py`
  - `ec_recovery_helper.py`
  - `load_wamr_stns.py`
  - `moti_infill_download.py`
  - `moti_infill_insert.py`
  - `moti_insert_files.py`


pyproject.toml
```diff
@@ -0,0 +1,68 @@
...
+[project]
+name = "crmprtd"
+version = "4.1.0"
+description = "Utility to download Environment Canada MPO-XML and insert it into PCIC's database"
+keywords = ["sql", "database", "data", "science", "climate", "oceanography", "meteorology"]
+requires-python = ">=3.8"
+authors=[
+    {name = "James Hiebert", email = "hiebert@uvic.ca"},
+    {name = "Rod Glover", email = "rglover@uvic.ca"},
+    {name = "Nic Rados"}
```
jameshiebert, line 31: Nik

pyproject.toml
```diff
@@ -0,0 +1,68 @@
...
+    "data/*.yaml",
+    "data/*.xsl",
+    "tests/data/*",
+    "wamr/station_substitutions.yaml",
+]
+
+[project]
+name = "crmprtd"
+version = "4.1.0"
+description = "Utility to download Environment Canada MPO-XML and insert it into PCIC's database"
```
jameshiebert, line 25: Could update this description, since the utility is now broader than this.

jameshiebert (changes requested):
Looks mostly good, but I think that you'll need to make some changes to the PyPI Publish action in `.github/workflows/pypi-publish.yml`. It still uses Python 3.6 and invokes `setup.py`. You'll probably want to temporarily change the "on: push: tags" so that you can test it before the next numbered release.

rod-glover:
Good catch, James! I published a test version, numbered 4.1.0a, to our pypi server, then installed it using pip only in a fresh virtual env. Install looked OK: `crmprtd_*` scripts were present and apparently working; contents of `venv/lib/python3.8/site-packages/` is as expected.

I'll remove the test setup from pypi-publish once you approve the PR.

jameshiebert (approved):
Looks great!

PR #142, opened 2023-02-01
Fix problems for actual use

### LINKED ISSUE #141: Problems in actual use
- Default cache and log file names contain "~", which is not resolved (i.e., just treated as a non-special filepath character) when opening a file from Python. It's Linux that resolves it when using it on the command line, which is not happening here.
- Env vars (e.g., PATH) are not passed through to subprocesses, either in Python or in Linux. 
  - E.g., in Linux, in `XXX=xxx foo.sh | bar.sh` only `foo.sh` has `XXX` defined. A similar thing is true for Python suprocesses.run. For the latter, there is a simple way of forwarding the env vars, but it has to be explicitly done, and isn't (yet).
  - `cron`, however, seems to behave differently. I can demonstrate (on my Ubuntu 20.04 instance) that all subprocesses spawned by any script it invokes get all the env variables defined in the `crontab`. 
  - Further testing will show whether this is true of variables defined on individual cron job lines (e.g., `* * * * * XXX=xxx foo.sh`). But if this holds up it (a) is very convenient and (b) seems risky/fragile. It may be a bad idea to rely on it, unless it is a specified feature.
- It would help to have a do-nothing standard "test" network so we can invoke it without, uh, doing anything.
- The remaining scripts in `crmprtd/execution/` are not included in `pyproject.toml`

rod-glover:
Resolves #141 

This PR:
- Uses `shell=True` in `subprocess.run`, which amongst other advantages, interprets bash symbols such as `~` in filepaths in the commands.
- Adds a do-nothing network named "_test" for live testing without actually downloading or processing anything.
- Adds the missing scripts from `scripts/` and `crmprtd/execution/`. (But uses a deprecated feature; choose your evil.)


rod-glover:
@jameshiebert , this is mostly boring, but if you care to cast your eyes over it. Or not.

PR #158, opened 2023-04-18
Completes BC Hydro data feed support

### LINKED ISSUE #40: The align phase needs to use caching from the database
In the `dev` branch, the align phase goes and does a database lookup for every single variable and every single station that it receives in a obs_tuple. When it get's thousands of obs_tuples this is *really slow*!

In the `master` branch, in the [`wamr` module](https://github.com/pacificclimate/crmprtd/blob/master/crmprtd/wamr.py) we had a strategy to deal with this by creating a cache mapping of both [variables](https://github.com/pacificclimate/crmprtd/blob/master/crmprtd/wamr.py#L48) and [histories](https://github.com/pacificclimate/crmprtd/blob/master/crmprtd/wamr.py#L28) at the beginning of the process, so that we only had to go to the database once, and then we could rely on the cache for the rest of the process.

We should implement this for align as well. It would save a *ton* of time.

jameshiebert:
We've been running this branch in production for the BC Hydro cache processing for a few weeks, so it's time to get it integrated into the main line and use it operationally.

This PR completes (near as I can tell) support for BC Hydro's sftp feed and the processing of the data it contains.

The bulk of the `crmprtd.bc_hydro.normalize` module [was introduced prior to v4.0.0](https://github.com/pacificclimate/crmprtd/commit/329cccfc) and as such, won't show up in this PR. Notable changes here include:

* Use of local memoization for database queries in the `align` phase. BCH had so many high-frequency, overlapping observations over a long during, that this was necessary to make everything run in a reasonable amount of time. Reducing round-trips to the database has been on our TODO list for a while, so win/win.
* Expands the download match pattern to include hourly/quarter-hourly data
* Adds a variable substitution table, since (the same) variables are named differently in different feeds.
* Fixes the use of strings vs. bytes in the `normalize` modules.

Closes #40


crmprtd/align.py
```diff
@@ -40,6 +41,29 @@
...
+                if not obj:
+                    return None
+                cache[key] = SimpleNamespace(
+                    **{attr: getattr(obj, attr) for attr in attrs}
+                )
+            return cache[key]
+
+        return memoize
+
+    return wrapper
```
rod-glover, line 64: Cool. I assume this is to support caching on only selected args/kwargs, unlike, say, `lrucache`, which caches on all args.

Also it looks like you use `SimpleNamespace` to imitate the behaviour of a SQLAlchemy query result (row). So maybe change the function name to `cached_query`?

And add a docstring describing this?

rod-glover: I'll remember `SimpleNamespace`, very handy.

crmprtd/align.py
```diff
@@ -383,15 +409,18 @@ def align(sesh, row, diagnostic=False):
             row.network_name,
         )
         return None
+    else:
+        var_unit = variable.unit
+        var_id = variable.id
```
rod-glover, line 414: :+1:

crmprtd/networks/bc_hydro/normalize.py
```diff
@@ -19,12 +22,25 @@ def normalize(file_stream):
...
     # mixture of a tab and other whitespace to make it fixed
     # width. This pattern encapsulates either
-    sep_pattern = re.compile(b"\\s*[\t,]+\\s*")
+    sep_pattern = re.compile("\\s*[\t,]+\\s*")
 
     # Detect if numbers are convertible
-    num_pattern = re.compile(rb"-?\d+(\.\d+)?$")
+    num_pattern = re.compile(r"-?\d+(\.\d+)?$")
+
+    variable_substitutions_path = "networks/bc_hydro/variable_substitutions.yaml"
```
rod-glover, line 30: Hmm, hardcoding makes me itchy, but it seems like this is a case for it. Unless we want to merge all variable substitutions in a single global substitutions file, which is probably not worth the trouble.

crmprtd/process.py
```diff
@@ -56,12 +56,14 @@ def add_process_args(parser):  # pragma: no cover
         "--start_date",
         help="Optional start time to use for processing "
         "(interpreted with dateutil.parser.parse).",
+        default=str(datetime.min),
```
rod-glover, line 59: :100:

crmprtd/process.py
```diff
@@ -112,9 +114,11 @@ def process(
     download_stream = sys.stdin.buffer
     norm_mod = get_normalization_module(network)
 
-    # The normalizer returns a generator that yields `Row`s. Convert to a list of Rows.
+    # The normalizer returns a generator that yields `Row`s. Convert to a set of Rows.
     log.info("Normalize: start")
-    rows = list(norm_mod.normalize(download_stream))
+
+    rows = {row for row in norm_mod.normalize(download_stream)}
+    print(f"Found {len(rows)} rows.")
```
rod-glover, line 121: Remove debug print?

tests/test_process.py
```diff
@@ -51,6 +72,10 @@ def num_observations_in_db(session):
 def test_process_by_date(
     start_date, end_date, expected_num_inserts, monkeypatch, caplog, crmp_session
 ):
+    importlib.reload(crmprtd.process)
+    importlib.reload(crmprtd.infer)
+    importlib.reload(crmprtd.align)
+
```
rod-glover, line 78: Suggest copying explanatory comments here too.

rod-glover (approved):
This looks great. Some nice code improvements, too.

crmprtd/align.py
jameshiebert: Yeah, that would have been easier to write the docstring back when I remembered how this worked! :stuck_out_tongue:

crmprtd/networks/bc_hydro/normalize.py
jameshiebert: Agreed in principle, but in practice, these things change very rarely. The other hard coded substitution that we have `crmprtd/networks/wamr/station_substitutions.yaml` has never changed.

PR #164, opened 2023-04-25
Filter out redundant observations before trying to insert (focus on network bc_hydro)

### LINKED ISSUE #163: Filter out redundant observations before trying to insert (focus on network bc_hydro)
This fix is primarily to address a problem that crops up with network `bc_hydro`. From the comments in `crmprtd/networks/bc_hydro/download.py`:

```
BC Hydro posts a rolling window (3 months) observing hourly data
once a week. Data is in txt files.

It is recommended to run this script once a week to collect
updated available data, supplemented with monitoring and alerting
for errors. If the script is run less than once every 3 months
you will miss data.
```

These comments were added in 2020. It might be worth checking or whether the terms have changed.

Let us parse the comments carefully. This is what I  get:
- Once a week, BCH posts updated data. (What day of week?)
- That data is for hourly observations.
- The download covers 3 month rolling window. When a new week's worth of data becomes available, all but that week (plus publication lag) have already been posted in previous weeks. This means that roughly 11/12 of the data is redundant for any given download.

We process BCH _daily_. That means any given download is mostly, and usually _entirely_ redundant if the database is up to date. Yikes. The problem this issue intends to correct is that processing such a large (~1.3 observations per download) with such high redundancy is very very very slow, and causes the server to crash. 

What is a suitable time period to filter on? 
- The time filtering must reach back at least one week before the current date (i.e., the date this is being run), plus any publication lag (time between observation time and publication time).
- We want to balance efficiency against a reasonable guarantee that we won't lose much data. Efficiency suggests just a week plus publication lag. But if we miss a week, we want to catch up, and if we reach back only about a week, we will lose it. 
- Can we query the database for a suitable initial time? That would be best.

rod-glover:
This PR does two things:
- Adds time-filtering for network bc_hydro:
  - When start time argument is not specified, now uses a default value for start time that is computed as (latest observation time for bc_hydro) - (publication lag).
  - See #163 for details on publication lag.
  - Publication lags are configured from a data file (`crmprtd/data/publication_lags.yaml`) and can be overridden by an argument.
- Eliminates a step from, and, I believe, possibly speeds up single-observation insertions.
  - Originally, there were two separately launched queries: "does this observation already exist; if not, insert it". 
  - Replaces it with the same one-step method used for bulk insertions, which handles the case that the uniqueness constraint rejects an already-existing query. 
  - This also DRYs up the insert code.
  - UPDATE: A test run on crmprtd today (May 1) shows that insertions for single observations are taking an average of ~500 ms, which is much faster than the observed time of > 3 s for single insertions in the previous version. Since both strategies, single-insert and bisection, share the same code, this demonstrates that we have improved the speed for single-insert strategy significantly.

Resolves #163

### DIFF
--- a/crmprtd/insert.py
+++ b/crmprtd/insert.py
@@ -127,7 +127,7 @@ def insert_single_obs(sesh, obs):
 def single_insert_strategy(sesh, observations):
     dbm = DBMetrics(0, 0, 0)
     for obs in observations:
-        insert_single_obs(sesh, obs)
+        dbm += insert_single_obs(sesh, obs)
     return dbm
 
 
@@ -160,7 +160,7 @@ def bisect_insert_strategy(sesh, observations):
     if len(observations) < 1:
         return DBMetrics(0, 0, 0)
     elif len(observations) == 1:
-        insert_single_obs(sesh, observations[0])
+        return insert_single_obs(sesh, observations[0])
     # The happy case: add everything at once
     else:
         try:
--- a/crmprtd/insert.py
+++ b/crmprtd/insert.py
@@ -95,27 +95,39 @@ def contains_all_duplicates(sesh, observations, sample_size):
     return True
 
 
-def single_insert_obs(sesh, obs):
-    dbm = DBMetrics(0, 0, 0)
-    for o in obs:
-        if obs_exist(sesh, o.history_id, o.vars_id, o.time):
-            dbm.skips += 1
-            log.debug("Observation already exists in database", extra={"obs_id": o.id})
-            continue
+def insert_single_obs(sesh, obs):
+    """Insert a single observation"""
+    try:
+        # Create a nested SAVEPOINT context manager to rollback to in the
+        # event of unique constraint errors
+        log.debug("New SAVEPOINT for single observation")
+        with sesh.begin_nested():
+            sesh.add(obs)
+    except IntegrityError as e:
+        log.debug(
+            "Failure, observation already exists",
+            extra={"observation": obs, "exception": e},
+        )
+        db_metrics = DBMetrics(0, 1, 0)
+    except InsertionError as e:
+        # TODO: InsertionError is an defined by crmprtd. It can't be raised by
+        #   SQLAlchemy unless something very tricky is going on. Why is this here?
+        log.warning(
+            "Failure occured during insertion",
+            extra={"observation": obs, "exception": e},
+        )
+        db_metrics = DBMetrics(0, 0, 1)
+    else:
+        log.info("Successfully inserted observations: 1")
+        db_metrics = DBMetrics(1, 0, 0)
+    sesh.commit()
+    return db_metrics
 
-        # value does not exist in obs_raw, continue with insertion
-        try:
-            sesh.add(o)
-            sesh.commit()
-            log.debug("Successfully inserted observation")
-            dbm.successes += 1
-        except Exception as e:
-            log.warning("Failure, an error occured.", extra={"e": e})
-            sesh.rollback()
-            dbm.failures += 1
-            raise InsertionError(
-                obs_time=o.time, datum=o.datum, vars_id=o.vars_id, hid=o.history_id, e=e
-            )
+
+def single_insert_strategy(sesh, observations):
+    dbm = DBMetrics(0, 0, 0)
+    for obs in observations:
+        insert_single_obs(sesh, obs)
     return dbm
 
 
@@ -129,7 +141,7 @@ def split(tuple_):
         return (tuple_[:i], tuple_[i:])
 
 
-def bisect_insert_strategy(sesh, obs):
+def bisect_insert_strategy(sesh, observations):
     """This function implements a recursive Obs insert strategy to
     handle unique constraint errors on members of the set. The
     strategy used is to optimistically attempt to insert the entire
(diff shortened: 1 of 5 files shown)


rod-glover:
Ran this on my dev laptop, with a local copy of a complete download dated Mar 3. The Mar 3 data is well outside the filtering period, and contains about 3M lines. Results were encouraging in terms of speed:
- Database query: 1:01
- Normalization: 37:00
- Align + filter: 0:01
- Insert: 0

 ```
{"asctime": "2023-04-27 22:22:27,112", "levelname": "WARNING", "name": "crmprtd.download_utils", "message": "Parameter end date 'None' is undefined or unparseable. Using the default '9999-12-31 23:59:59'"}
{"asctime": "2023-04-27 22:22:27,115", "levelname": "INFO", "name": "crmprtd.process", "message": "Querying database for latest obs time..."}
{"asctime": "2023-04-27 22:23:36,762", "levelname": "INFO", "name": "crmprtd.process", "message": "For network bc_hydro, latest obs time from query = 2023-04-18 12:15:00"}
{"asctime": "2023-04-27 22:23:36,763", "levelname": "INFO", "name": "crmprtd.process", "message": "Final time filter parameters: start_date = 2023-04-04 12:15:00+00:00, end_date = 9999-12-31 23:59:59.999999+00:00"}
{"asctime": "2023-04-27 22:23:36,767", "levelname": "INFO", "name": "crmprtd.process", "message": "Normalize: start"}
{"asctime": "2023-04-27 23:00:13,443", "levelname": "INFO", "name": "crmprtd.process", "message": "Normalize: done"}
{"asctime": "2023-04-27 23:00:13,443", "levelname": "INFO", "name": "crmprtd.process", "message": "Align + filter: start"}
{"asctime": "2023-04-27 23:00:14,513", "levelname": "INFO", "name": "crmprtd.process", "message": "Align + filter: done"}
{"asctime": "2023-04-27 23:00:14,513", "levelname": "INFO", "name": "crmprtd.process", "message": "Count of observations to process: 0"}
{"asctime": "2023-04-27 23:00:14,513", "levelname": "INFO", "name": "crmprtd.process", "message": "Insert: start"}
{"asctime": "2023-04-27 23:00:14,517", "levelname": "INFO", "name": "crmprtd.insert", "message": "Using Single Insert Strategy"}
{"asctime": "2023-04-27 23:00:14,518", "levelname": "INFO", "name": "crmprtd.insert", "message": "Data insertion complete"}
{"asctime": "2023-04-27 23:00:14,518", "levelname": "INFO", "name": "crmprtd.process", "message": "Insert: done"}
{"asctime": "2023-04-27 23:00:14,518", "levelname": "INFO", "name": "crmprtd.process", "message": "Data insertion results", "results": {"successes": 0, "skips": 0, "failures": 0, "insertions_per_sec": 0.0}, "network": "bc_hydro"}
```

**Comment by rod-glover** [comment]
Version 4.3.1.dev2 is working its way through a download. Because the database is so far behind, it is processing all of the download, which means 
- 3,182,621 "rows" (which took about 7 minutes to normalize)
- 401,247 observations that survived the normalization process

The observations are being inserted at a rate of about 7600 obs/hr, which is extremely slow. Projected time to complete is about 52 hr. This shows why kicking off this pipeline every day soon choked the crmprtd server.

However, the progress so far shows that this version is working fine. On that basis, I propose to merge it.

**Comment by rod-glover** [comment]
Question: Why was normalization so fast today? The previous time, running on my laptop against a local copy of a full download, it took 37 minutes -- was that a mistaken calculation? (I don't think so, but it's possible.)

File: crmprtd/insert.py [hunk 1]
```diff
@@ -129,7 +141,7 @@ def split(tuple_):
         return (tuple_[:i], tuple_[i:])
 
 
-def bisect_insert_strategy(sesh, obs):
+def bisect_insert_strategy(sesh, observations):
```
jameshiebert, line 144: Beware of changing variable names, lest you forever be known (i.e. `git blame`d) as the author of this code.

crmprtd/process.py
```diff
@@ -57,14 +56,26 @@ def add_process_args(parser):  # pragma: no cover
...
-        default=str(datetime.min),
     )
     parser.add_argument(
         "-E",
         "--end_date",
         help="Optional end time to use for processing "
         "(interpreted with dateutil.parser.parse).",
-        default=str(datetime.max),
+    )
+    parser.add_argument(
```
jameshiebert, line 66: Good idea, though I don't think that you have clearly articulated in the help strings how this option interacts with `start_date` and `end_date` and which takes precedence. The somewhat convoluted logic below, is indicative of that lack of clarity.

crmprtd/process.py
```diff
@@ -111,21 +123,72 @@ def process(
...
+    # This query typically takes on the order of 90 s, but that cost is far lower than
+    # that of trying to insert - one by one - observations that are already present.
+    # A publication lag (days) is subtracted from the latest obs time to determine
+    # the start_date.
+
+    with resource_stream("crmprtd", "data/publication_lags.yaml") as pl_f:
+        publication_lags = yaml.safe_load(pl_f)
+
+    if network in publication_lags and start_date is None:
+        log.info(f"Querying database for latest obs time for network {network}...")
```
jameshiebert, line 142: I find this logic very dangerous and do not like it. It's not uncommon for networks to add observations to their archives in a non-monotonically increasing fashion. Filtering obs that are before the most recent obs_time doesn't give the opportunity to catch updates from the network, and if I'm not mistaken, it will completely break any of our use cases where we infill gaps (say if our `crmprtd` VM goes down overnight and is restarted later).

crmprtd/process.py
```diff
@@ -111,21 +123,72 @@ def process(
...
+        log.warning(
+            f"Publication lag (value {publication_lag}) is specified for "
+            f"network {network}, which does not use it."
+        )
+
+    # For all other cases, replace start_date == None with datetime.min.
+    if start_date is None:
+        start_date = datetime.min
+
+    # At present, this is impossible, but hey, it's cheap insurance.
```
jameshiebert, line 171: Not sure what you mean by this comment. What is impossible?

crmprtd/process.py
```diff
@@ -111,21 +123,72 @@ def process(
...
     # Get the normalizer for the specified network.
     download_stream = sys.stdin.buffer
     norm_mod = get_normalization_module(network)
 
-    # The normalizer returns a generator that yields `Row`s. Convert to a set of Rows.
     log.info("Normalize: start")
-
+    # The normalizer returns a generator that yields `Row`s. Convert to a set of Rows.
+    # It is probably better to use a dict for this to preserve order.
+    # See https://stackoverflow.com/a/9792680
```
jameshiebert, line 187: This SO comment is very out of date. I looked into this (should have noted it in the comment, b/c I can't find the evidence now), but contemporary Python sets *do* preserve insertion order.

tests/test_insert.py
```diff
@@ -11,7 +11,8 @@
     get_sample_indices,
     obs_exist,
     contains_all_duplicates,
-    single_insert_obs,
+    contains_all_duplicates,
```
jameshiebert, line 14: Why are there 2 `contains_all_duplicates`?

jameshiebert (changes requested):
I think that the changes to the date filtering logic in process warrant some more discussion. If I'm reading them correctly, they have wide ranging (and non-ideal) implications. See the comments for more details.

crmprtd/process.py
jameshiebert: Hmmm... hold the presses on that... I may be incorrect here.

crmprtd/process.py
rod-glover: I understand and I was not sure this was a good idea either. I'm glad to have this feedback.

Motivation was basically that it takes (as of this PR) a very long time to process an entire BCH download file, because of the insert strategy/ies presently in use. (Improvements to the insertion method mean this may no longer be an issue.) I knew that just using the latest obs time in the database was naive, even assuming the database was up to date with all previous downloads. The publication lag was intended to backstop this, while still allowing us to process only a subset of the records in the download. I recognize the reasoning here may be flawed.

This only applies when the start time is not specified. This does "break" the previous usage: previously leaving it unspecified meant "process all". Now we would have to specify it explicitly.

This whole explanation makes me regret this approach.

crmprtd/process.py
rod-glover: Meaning: The script at present will never pass in a None value.

tests/test_insert.py
rod-glover: Goodness knows. Removed.

crmprtd/insert.py
rod-glover: That does mean you are forever stuck with bad things like confusing variable names.

crmprtd/process.py
jameshiebert: Ah OK, that makes sense.

rod-glover:
Superseded by #168 . crmprtd is now sufficiently fast that filtering on time is no longer necessary.

PR #168, opened 2023-05-04
Improve align and insert phase performance

### LINKED ISSUE #165: Improve insert performance for observations
We have a problem with bulk inserts of observations: There may be many redundant (already-existing) records which cause any given single insert to fail.

We currently have two approaches to this in the Python code: For blocks of data which appear to have few redundant records, we use a bisection strategy, and for blocks which have many redundant records, we use a single-insert strategy. The latter is particularly slow for reasons not understood. It seems possible that we could implement one or both of these as a stored procedure, which might reduce whatever overhead is involved in an insert (whether single or multiple records) query.

Possible better alternatives:
- Understand better what causes single insertions (or possibly any single query) to be slow.
- Use a different strategy in our Python query formulations to allow for inserts that violate the uniqueness constraint on the observation table without resorting to the single-insert strategy (or indeed the bisection strategy).
- Create a stored procedure that implements insert with fails internally.

rod-glover:
Resolves #165 

This PR:
- Adds a BULK insert strategy that uses the PG `INSERT .... ON CONFLICT DO NOTHING` feature to insert large numbers of rows into the Obs table, some of which may violate the unique constraint on that table. Duplicate rows are ignored, the rest are inserted. It's fast.
- Preserves the old strategy(ies), under the rubrics SINGLE, CHUNK_BISECT, and ADAPTIVE (the original insert strategy).
- Improves slightly the behaviour of the single-insert function that is resorted to in the worst case. It's still quite slow.
- Allows the user to pick the strategy with the -R/--insert_strategy (-i, -I, S, and may other likely option flags were all taken, so R for inseRt !). Default is BULK.
- Substantially improves the speed of the align phase, by caching None results from the database as well as non-None results. That results in an approximately 50x speedup in the align phase (~30-40 s vs. ~1 hr for ~1 M rows).
- Adds progress logging for the align phase, which was useful for discovering the None/not-None improvement, and which remains interesting and unobtrusive now.

@jameshiebert , if it seems that you have seen some of this code before, you have. I cherry-picked useful commits from #164 (which was most of them) and added the bulk insert feature on top of them. The dangerous time filtering algo is of course not present in this version; all the time filtering code is left unchanged from the mainline. I think it no longer matters if we try to insert large numbers of already-existing records if BULK is used.

### DIFF
--- a/crmprtd/align.py
+++ b/crmprtd/align.py
@@ -442,12 +442,8 @@ def align(sesh, row, diagnostic=False):
         )
         return None
 
-    # To allow the use of the PostgreSQL dialect-specific `insert` function, we return a
-    # dict with the relevant key-value mapping. This may need to be converted to an Obs
-    # object, or not, depending on how it is used.
-    return {
-        "history_id": history.id,
-        "obs_time": row.time,
-        "datum": datum,
-        "vars_id": var_id,
-    }
+    # Create Obs object.
+    # Note: We are very specifically creating the Obs object here using the ids
+    # to avoid SQLAlchemy adding this object to the session as part of its
+    # cascading backref behaviour https://goo.gl/Lchhv6
+    return Obs(history_id=history.id, time=row.time, datum=datum, vars_id=var_id)
--- a/crmprtd/align.py
+++ b/crmprtd/align.py
@@ -446,4 +446,13 @@ def align(sesh, row, diagnostic=False):
     # Note: We are very specifically creating the Obs object here using the ids
     # to avoid SQLAlchemy adding this object to the session as part of its
     # cascading backref behaviour https://goo.gl/Lchhv6
-    return Obs(history_id=history.id, time=row.time, datum=datum, vars_id=var_id)
+    # return Obs(history_id=history.id, time=row.time, datum=datum, vars_id=var_id)
+
+    # To allow the use of the PostgreSQL dialect-specific `insert` function, we return a
+    # dict with the relevant key-value mapping. And native table key names.
+    return {
+        "history_id": history.id,
+        "obs_time": row.time,
+        "datum": datum,
+        "vars_id": var_id,
+    }
--- a/crmprtd/align.py
+++ b/crmprtd/align.py
@@ -67,7 +67,7 @@ def cached_function(attrs):
             key = (args) + tuple(kwargs.items())
             if key not in cache:
                 obj = f(sesh, *args, **kwargs)
-                log.info(f"Cache miss: {f.__name__} {key} -> {repr(obj)}")
+                log.debug(f"Cache miss: {f.__name__} {key} -> {repr(obj)}")
                 cache[key] = obj and SimpleNamespace(
                     **{attr: getattr(obj, attr) for attr in attrs}
                 )
--- a/crmprtd/align.py
+++ b/crmprtd/align.py
@@ -42,9 +42,9 @@ for def_ in (
 
 
 def cached_function(attrs):
-    """A context manager that can be used to cache database results
+    """A decorator factory that can be used to cache database results
 
-        Neither database sessions (i.e. the sesh parameter of each wrapped
+    Neither database sessions (i.e. the sesh parameter of each wrapped
     function) nor SQLAlchemy mapped objects (the results of queries) are
     cachable or reusable. Therefore one cannot memoize database query
     functions using builtin things like the lrucache.
@@ -55,8 +55,8 @@ def cached_function(attrs):
     and c) accepting as a parameter a list of attributes to retrieve and
     store in the cache result.
 
-        args (except sesh) and kwargs to the wrapped function are used as
-        the cache key, and results are the parametrized object attributes.
+    args (except sesh) and kwargs to the wrapped function are used as
+    the cache key, and results are the parametrized object attributes.
     """
 
     def wrapper(f):
@@ -67,9 +67,8 @@ def cached_function(attrs):
             key = (args) + tuple(kwargs.items())
             if key not in cache:
                 obj = f(sesh, *args, **kwargs)
-                if not obj:
-                    return None
-                cache[key] = SimpleNamespace(
+                log.info(f"Cache miss: {f.__name__} {key} -> {repr(obj)}")
+                cache[key] = obj and SimpleNamespace(
                     **{attr: getattr(obj, attr) for attr in attrs}
                 )
             return cache[key]
--- a/crmprtd/align.py
+++ b/crmprtd/align.py
@@ -442,14 +442,9 @@ def align(sesh, row, diagnostic=False):
         )
         return None
 
-    # Create Obs object.
-    # Note: We are very specifically creating the Obs object here using the ids
-    # to avoid SQLAlchemy adding this object to the session as part of its
-    # cascading backref behaviour https://goo.gl/Lchhv6
-    # return Obs(history_id=history.id, time=row.time, datum=datum, vars_id=var_id)
-
     # To allow the use of the PostgreSQL dialect-specific `insert` function, we return a
-    # dict with the relevant key-value mapping. And native table key names.
+    # dict with the relevant key-value mapping. This may need to be converted to an Obs
+    # object, or not, depending on how it is used.
     return {
         "history_id": history.id,
         "obs_time": row.time,
--- a/crmprtd/insert.py
+++ b/crmprtd/insert.py
@@ -217,6 +217,18 @@ def bisect_insert_strategy(sesh, observations):
         return db_metrics
 
 
+def obs_to_pg_insert_dict(obs):
+    """
+    Convert an Obs object to a dict suitable for consumption by pg_insert.
+    """
+    return {
+        "history_id": obs.history_id,
+        "obs_time": obs.time,
+        "datum": obs.datum,
+        "vars_id": obs.vars_id,
+    }
+
+
 def insert_bulk_obs(sesh, observations):
     """
     This method performs a bulk insert of observations using the PostgreSQL dialect
@@ -241,11 +253,11 @@ def insert_bulk_obs(sesh, observations):
         with sesh.begin_nested():
             result = sesh.execute(
                 pg_insert(Obs)
-                .values(observations)
+                .values([obs_to_pg_insert_dict(o) for o in observations])
                 .on_conflict_do_nothing()
                 .returning(Obs.id)
             ).fetchall()
-    except DBAPIError:
+    except DBAPIError as e:
         # Something really unanticipated happened. Duplicate rows do not trigger an
         # exception.
         log.exception("Unexpected error during bulk insertion")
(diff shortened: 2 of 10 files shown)


crmprtd/insert.py
```diff
@@ -142,85 +178,133 @@ def bisect_insert_strategy(sesh, obs):
...
+    """
+    This method performs a bulk insert of observations using the PostgreSQL dialect
+    INSERT ... ON CONFLICT DO IGNORE clause that allows bulk inserts with duplicates
+    to proceed without raising an exception (dups are ignored; new rows are inserted).
+
+    NOTE: This only works for PostgreSQL databases. Other dialects may support a
+    similar operation, but this code does not handle those cases.
+
+    :param sesh: SQLAlchemy database session
+    :param observations: List of observations to insert
```
jameshiebert, line 230: Technically, it's a list of `dict`s, right? I think that supports my contention that the change in interface between `align` and `insert` is confusing. By only *reading* the code, it's hard to tell what's an actual `Obs` object, and what are `dict`s.

crmprtd/insert.py
```diff
@@ -142,85 +178,133 @@ def bisect_insert_strategy(sesh, obs):
...
+    # Avoid that.
+    if num_to_insert == 0:
+        return DBMetrics(0, 0, 0)
+
+    try:
+        with sesh.begin_nested():
+            result = sesh.execute(
+                pg_insert(Obs)
+                .values(observations)
+                .on_conflict_do_nothing()
```
jameshiebert, line 245: Since we're not using the ORM mappings here for the insert, I have to wonder whether a significant part of the optimization is simply in the bypassing of the ORM and using raw SQL. E.g. when we create an instance of `Obs`, does that require a round trip (or multiple round trips) to the database to fill out the object attributes? If so, that would be the slow part, regardless of how we try to do the insert.

crmprtd/insert.py
```diff
@@ -142,85 +178,133 @@ def bisect_insert_strategy(sesh, obs):
...
+            extra={"num_obs": num_inserted},
+        )
+    sesh.commit()
+    return DBMetrics(num_inserted, num_to_insert - num_inserted, 0)
+
+
+def insert(
+    sesh,
+    observations,
+    strategy=InsertStrategy.BULK,
```
jameshiebert, line 283: Part of what I liked about the old insert code was that it was adaptive and didn't require user control (only user *tuning*) over which insert strategy was used. If it detected a large percentage of duplicates, single insert was faster, so it would use that. Otherwise it would use the bulk/bisect strategy.

So that raises the question to me as to whether there are any problematic scenarios when your new bulk insert strategy *wouldn't* be faster. I suspect not, in which case the adaptiveness is now irrelevant, so making it user specified (with a sane default) is fine. But do you see any potential pitfalls?

That said, the bisect insert strategy and the single insert strategy are separate and you have conflated them into a hybrid of select-both-and-get-one-or-the-other. If we're going to commit to requiring user selection, I'd add single-insert as an option and just pull out the adaptive sampling code.

crmprtd/align.py
```diff
@@ -442,8 +442,12 @@ def align(sesh, row, diagnostic=False):
...
-    return Obs(history_id=history.id, time=row.time, datum=datum, vars_id=var_id)
+    # To allow the use of the PostgreSQL dialect-specific `insert` function, we return a
+    # dict with the relevant key-value mapping. This may need to be converted to an Obs
+    # object, or not, depending on how it is used.
+    return {
+        "history_id": history.id,
+        "obs_time": row.time,
+        "datum": datum,
+        "vars_id": var_id,
+    }
```
jameshiebert, line 453: I don't like this change of interface between `align` and `insert` for a couple reasons.

1. To me, it feels like a leaky abstraction. We want a clean interface between the `align` and `insert` phases, but then the fact that we still do more conversions on the data after they go to `insert` (or not) feels... messy to me. Though I'm having trouble articulating why its necessarily problematic. I guess to me the "promise" of `align` has always been, "I'm passing you something that theoretically should be able to go into the database", and a `dict` feels much, much less guaranteed.
2. It's loosely specified. "a dict" can be or contain anything. It's kind of like saying: "give me your data in JSON and everything will work!" (but it doesn't). This can be problematic both for the programmer ("what does the dict need to contain?") and for the Python interpreter which *could* adaptively optimize for a more tightly constrained data structure, but cannot for a loosely and openly defined `dict`. For most things the performance may not matter, but for the central data being passed between pieces of the program, it may.

If we think that we want to not use the ORM's `Obs` class, than maybe we should at least create a [`NamedTuple`](https://docs.python.org/3/library/typing.html#typing.NamedTuple), or some other more specified and immutable object.

jameshiebert (changes requested):
Despite my comments and discussion, I do really like this PR and appreciate the ultimate *results* of it ("MOAR PERFORMANCE!"). Thanks! I think there are a few tweaks that will help maintain the spirit of how the system was originally designed.

crmprtd/insert.py
```diff
@@ -142,85 +178,133 @@ def bisect_insert_strategy(sesh, obs):
...
+):
     """
     Insert a collection of observations.
 
     :param sesh: SQLAlchemy database session
-    :param observations: Obs objects to insert
-    :param sample_size: Size of sample of observations to use to test if all duplicates.
+    :param observations: For BULK insert strategy, list of dicts containing Obs rows
+        to insert. For OTHER insert strategy, list of Obs objects to insert.
+    :param strategy: Insert strategy. BULK is currently fastest.
```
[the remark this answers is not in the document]
rod-glover: Oops, that should have been BISECTION. The list of options is defined by the enum InsertStrategy. I don't think it's a good idea to repeat it here, since if it changes this derived string would have to change. I'll put in a note "see InsertStrategy".

crmprtd/insert.py
rod-glover: Off the cuff, I don't think so. This is no more "raw SQL" than any other use of SQLAlchemy.

Not sure precisely what you mean by "create an instance of `Obs`". If you mean _within the code_, then it takes _zero_ round trips. If you mean _when an object is inserted in the database_, then it takes one round trip, as any other query would. Unless you emit a query, the database is not involved.

crmprtd/insert.py
rod-glover: I thought about this, and decided not to mess with any of the existing code, apart from making it optional. My mistake here was naming the alternative "BISECTION". It should have remained "OTHER" as it was earlier -- but that seemed too vague.

I haven't been able to foresee any circumstances under which the new bulk insert strategy would be slower. It requires precisely one round trip per set of rows to insert. Single insert and bisection all require more than one, except in the case that a large block of bisectable observations inserts successfully the first time. There's evidently a significant overhead to a round trip (and I don't know why).

Given that adaptive inserting avoids the worts case in which bisection bisects continually down to single inserts (which would require _log n_ insert attempts before succeeding or failing on a single insert), I don't think this it's a good idea to separate them ... unless for purposes of controlling comparisons, which might not be a bad idea. How about this:
- BULK (or some better name): INSERT ... ON CONFLICT DO NOTHING
- SINGLE: single insert
- BISECTION: ...
- ADAPTIVE: the old adaptive way, currently selected by BISECTION (bad name).

crmprtd/align.py
rod-glover: I thought about this. I considered both your objection/solution and the inverse of what's here (i.e., return Obs, and add a conversion from Obs to a dict for the bulk insertion with pg_insert). I prefer the second; it's cleaner and more in the spirit of SQLAlchemy. All I can say is I was working quickly ...

crmprtd/insert.py
```diff
@@ -142,85 +178,133 @@ def bisect_insert_strategy(sesh, obs):
...
+            result = sesh.execute(
+                pg_insert(Obs)
+                .values(observations)
+                .on_conflict_do_nothing()
+                .returning(Obs.id)
+            ).fetchall()
+    except DBAPIError as e:
+        # Something really unanticipated happened. Duplicate rows do not trigger an
+        # exception.
+        log.warning(f"Unexpected error during bulk insertion: {e}")
```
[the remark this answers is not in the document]
rod-glover: Agreed. In that case, I'd use [`log.exception`](https://docs.python.org/3/library/logging.html#logging.exception).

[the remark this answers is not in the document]
jameshiebert: Doesn't `log.exception` actually raise the exception? Is that what you want?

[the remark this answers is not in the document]
rod-glover: Docs:

`logging.exception(msg, *args, **kwargs)`

Logs a message with level ERROR on the root logger. The arguments are interpreted as for [debug()](https://docs.python.org/3/library/logging.html#logging.debug). Exception info is added to the logging message. This function should only be called from an exception handler.

crmprtd/insert.py
rod-glover: Going ahead with this, which covers all bases. Can be modified later if necessary.

crmprtd/insert.py
rod-glover: Not any more.

rod-glover:
In a recent test run from a laptop workstation, targeting the production crmp database, reprocessing an entire BCH download file took 8.5 _minutes_. More detailed stats
- file contains 24,727,638 rows
- after normalization: 3,291,148 rows  (start: 11:06:34, end: 11:11:23, elapsed: ~5 min)
- after alignment and filtering: 1,161,830 observations to insert (start: 11:11:24, end: 11:11:58, elapsed: ~34 s)
- insertion (bulk strategy): start: 11:11:58, end: 11:14:59, elapsed: ~3 min
- total time: start: 11:06:34, end: 11:14:59, elapsed: ~8.5 min

The most recent processing of a BCH download file using the previous (now option ADAPTIVE) method took 5 _days_. That file, however, contained all new records, in contrast to the already-processed file used in this test, so the insertion time may be considerably greater in that case. Other tests do indicate an approximate 50x to 100x speedup over that method.

rod-glover:
@jameshiebert , ready for another review, thanks.

crmprtd/more_itertools.py
```diff
@@ -0,0 +1,65 @@
...
+
+    The logger is given by the argument `log`, which is typically a Python logger
+    function.
+
+    Interesting that the explanation is 3x as long as the code.
+    """
+    count = 0
+    when = tuple(when)
+    front, last = when[0 : len(when) - 1], when[len(when) - 1]
+    log(f"log_progress: when={when}, front={front}, last={last}")
```
rod-glover, line 55: Remove

crmprtd/insert.py
```diff
@@ -142,85 +187,157 @@ def bisect_insert_strategy(sesh, obs):
...
-def insert(sesh, observations, sample_size):
+def chunk_bisect_insert_strategy(sesh, observations):
+    log.info("Using Chunk + Bisection Strategy")
+    dbm = DBMetrics(0, 0, 0)
+    for chunk in bisection_chunks(observations):
+        dbm += bisect_insert_strategy(sesh, chunk)
+    return dbm
+
+
+def obs_to_pg_insert_dict(obs):
```
jameshiebert, line 229: Much better.

jameshiebert (commented):
Looks much better! Thanks for doing a second round of work on it!

crmprtd/insert.py
rod-glover: Stupid thing is that I knew I should do this all along....

jameshiebert (approved):

PR #176, opened 2023-07-31
i97 Insert observations grouped first by network

### LINKED ISSUE #97: Insert logging assumes homogenous networks
`crmprtd`'s `insert` module is set up to be able to take an arbitrary list of `Obs` objects and insert them all in a batch. So, theoretically, one could have several different networks represented among the set.

However, the [`DBMetrics`](https://github.com/pacificclimate/crmprtd/blob/master/crmprtd/insert.py#L24) does not track individual networks, and the [`process()` function](https://github.com/pacificclimate/crmprtd/blob/master/crmprtd/process.py#L47) assumes that all the `Obs` in the batch are of the same network [and logs it as such](https://github.com/pacificclimate/crmprtd/blob/master/crmprtd/process.py#L47).

The logging and analytics per network are very important, so we need to fix this and keep track of network-level information. We can probably have the `insert` function first sort the `Obs` into homogeneous groups, and then basically return the same reporting structure, but as a `dict` per network.

QSparks:
Added two functions to [insert.py](https://github.com/pacificclimate/crmprtd/blob/6f3bbcf8490efdb17367ee1a162285a3f925361f/crmprtd/insert.py#L57) that will sort an arbitrary list of observations into a dictionary where <network_name> is the key. This way network-level logging is maintained even if calls to insert have observations from multiple networks.

resolves #97

### DIFF
--- a/crmprtd/insert.py
+++ b/crmprtd/insert.py
@@ -407,7 +407,7 @@ def insert(
             "insertions_per_sec": round(dbm.successes / tmr.run_time, 2),
         }
         log.info(
-            "Insert for network: {network}: done".format(
+            "Insert for network {network}: done".format(
                 network=get_network_name(sesh, obs[0])
             )
         )


crmprtd/process.py
```diff
@@ -109,6 +111,22 @@ def get_normalization_module(network):
     return import_module(f"crmprtd.networks.{network}.normalize")
 
 
+@functools.lru_cache(maxsize=None)
+def get_network(sesh, var_id):
```
jameshiebert, line 115: The `sesh` variable isn't cacheable since it is an object of a database session which necessarily involves "state". I suspect that you were trying to do something similar to the `cached_function` decorator that you imported above (but don't actually use)?

If you choose to use `cached_function` here, I think that you should move it out of the `align` module to another module that is higher level and would be intended to be used across the entire code base.

crmprtd/process.py
```diff
@@ -196,16 +214,22 @@ def process(
...
-        sesh,
-        observations,
-        strategy=insert_strategy,
-        bulk_chunk_size=bulk_chunk_size,
-        sample_size=sample_size,
-    )
-    log.info("Insert: done")
-    log.info("Data insertion results", extra={"results": results, "network": network})
+
+    for network_key in obs_by_network_dict:
```
jameshiebert, line 221: What you've done here reminds me of the Python standard library function [`itertools.groupby`](https://docs.python.org/3/library/itertools.html#itertools.groupby) (which would be similar to the SQL `GROUP BY` clause). Do you think that there's a way to utilize `itertools.groupby` to streamline what you have written here?

jameshiebert (changes requested):
The overall structure of the approach that you've taken looks great, but I have a couple of questions/comments that I think need to be addressed.

If you're able to write some unit tests that exercise these new functions, I think that would be helpful, because my read of the `get_network()` function is that it will not work (w.r.t. the caching) as written.

crmprtd/insert.py
```diff
@@ -54,6 +54,22 @@ def max_power_of_two(num):
     return 2 ** floor(mathlog(num, 2))
 
 
+def get_network_name(sesh, obs):
+    Obs_var = sesh.query(Variable).filter_by(id=obs.vars_id).first()
+    return Obs_var.network.name
+
+
+def obs_by_network(observations, sesh):
+    observations.sort(key=lambda obs: get_network_name(sesh, obs))
```
jameshiebert, line 63: Good catch that the input to `groupby` needs to be sorted. I had missed that fact.

I'd recommend using the `sorted()` builtin function here for a few reasons:
1. It's often considered better practice to *not* modify input data in place, if possible.
2. `sorted()` is more flexible in that it can take an *iterable* and it doesn't rely on the input data being a list or dict or a specific object that has a `.sort` attribute.
3. The `sorted()` builtin has received a lot of developer attention from the core devs w.r.t. performance. There's a higher chance that it might do some optimizing behind the scenes that you don't need to worry about.

crmprtd/insert.py
```diff
@@ -54,6 +54,22 @@ def max_power_of_two(num):
...
+    obs_by_network_dict = {
+        network_name: list(obs_group)
+        for network_name, obs_group in groupby(
```
jameshiebert, line 66: Have you considered the pros/cons of simply returning the return value of `groupby()` here? Could you discuss them?

crmprtd/insert.py
```diff
@@ -54,6 +54,22 @@ def max_power_of_two(num):
     return 2 ** floor(mathlog(num, 2))
 
 
+def get_network_name(sesh, obs):
```
jameshiebert, line 57: This function is used *very* often. It's used in two tight loops that would iterate over all of the observations. Since it reaches out to the database, this is going to involve a lot of round trips (where are several orders of magnitude slower than anything happening in RAM).

I think that you'll *really* want to take a look at the `cached_function` decorator that's defined in `align.py`.

tests/test_insert.py
```diff
@@ -195,3 +198,44 @@ def test_single_insert_obs_not_unique(test_session):
...
+                time=datetime(2012, 9, 24, 6, tzinfo=pytz.utc),
+            )
+            obs.append(observation)
+
+    results = insert(test_session, obs)
+    assert networks_logged(obs, test_session, caplog)
+
+
+def networks_logged(obs, test_session, caplog):
+    obs_by_network_dict = obs_by_network(obs, test_session)
```
jameshiebert, line 233: I understand the motivation to reuse code here in the testing. You have a nice function that you've written to group the observations by network, so why not use it for the test? One reason *not* to, is that it essentially results in circular logic. You have a test that executes `obs_by_network()` and you compare that against results that *also* execute `obs_by_network()` :smile: It's like the reflexive property. Anything is always equal to itself.

I'd advise that you write a separate test *specifically* testing `obs_by_network()` under a few different scenarios to ensure that it's working in isolation.

Then `test_multi_networks` might become simpler, or at least you can trust that `obs_by_network()` is safe and your test of `insert()` should be checking other things.

tests/test_insert.py
```diff
@@ -195,3 +198,44 @@ def test_single_insert_obs_not_unique(test_session):
...
+            var = hist.station.network.variables[0]
+            observation = Obs(
+                history_id=hist.id,
+                datum=10,
+                vars_id=var.id,
+                time=datetime(2012, 9, 24, 6, tzinfo=pytz.utc),
+            )
+            obs.append(observation)
+
+    results = insert(test_session, obs)
```
jameshiebert, line 228: Is there any information contained in the `results` object that you could make `assert`ions about that would be valuable for testing?

tests/test_insert.py
```diff
@@ -195,3 +198,44 @@ def test_single_insert_obs_not_unique(test_session):
...
+    assert networks_logged(obs, test_session, caplog)
+
+
+def networks_logged(obs, test_session, caplog):
```
jameshiebert, line 232: If you rename this to be `networks_are_logged` it might read more clearly as to what it does.

And what about simply passing in a tuple of network names that you expect to be logged. It would simplify the signature:

```
def networks_are_logged(networks, caplog):
```

jameshiebert (changes requested):
It's looking much better! Good work :+1: 

A few more comments that I think will further improve the code.

crmprtd/insert.py
QSparks: From the [docs](https://docs.python.org/3/library/itertools.html#itertools.groupby):
“The returned group is itself an iterator that shares the underlying iterable with groupby(). Because the source is shared, when the groupby() object is advanced, the previous group is no longer visible. So, if that data is needed later, it should be stored as a list”

Since we are planning to pass lists of observations grouped by their associated network, using dictionary comprehension with network as the key and a list of observations as the value is preferred.

The pros that come to mind are conciseness and memory efficiency. The return value would be a k,v pair that gets immediately consumed. This means that once the observations associated with a key have been iterated we don’t store them in memory as we do with the dictionary approach.

jameshiebert (changes requested):
Looks much better! Last two comments and then I think we're done:
1. I think `log_helpers` as a separate module is a good idea for these two functions. However, the name is a bit of a misnomer, since `cached_function` isn't really used in logging. How about `db_helpers.py`?
2. I think that it's important for db_metrics to be returned (or `yield`ed) per network. Different networks have different performance characteristics. I'd like to you swap the loops there and have each per-network iteration `yield` the db metrics results. You may need to adjust the calling code in `process.py` to make it work.

PR #182, opened 2025-07-10
Update ec network normalize to support variable subsitutions

Nospamas:
Updated crmprtd based on bc_hydro network variable substitutions across to the ec network as well.

Current setup does the following variables:

```
'air_temperature_yesterday_high': 'air_temperature'
'air_temperature_yesterday_low': 'air_temperature'
'total_precipitation': 'total_precipitation'
'wind_direction': 'wind_from_direction'
'wind_gust_speed': 'wind_speed'
```

Some unknowns for me due to unfamiliarity with the project leaves me with the following questions:

- Will the duplicated air_temperature be an issue? Will cell method be correctly pulled?
- A number of the variables coming back from EC in the XML sheet should maybe be 0 not empty, specifically the precipitation related variables. This could do with a sanity check I think
- I've included total_precipitation in substitutions for completeness but it doesn't switch it to anything
- There remains a bug? with one of the tests (test_download_cache) related to timezones, it fails on my machine due to not being in UTC I think.


crmprtd/networks/ec/normalize.py
```diff
@@ -1,7 +1,41 @@
...
+from importlib.resources import files
+from crmprtd import Row
 from crmprtd.swob_ml import normalize as swob_ml_normalize
 
+log = logging.getLogger(__name__)
+
 
 def normalize(file_stream):
-    return swob_ml_normalize(
+
```
jameshiebert, line 12: I'm always skeptical when a patch adds LOCs to our repo. :) The approach you've taken here is reasonable, but I don't think that this is (or will be) specific to ECCC. Could you take what you've done here and incorporate it into the `swob_ml` module? Then if other networks change their variable names in the future, we will have an easy place to add variable mappings.

jameshiebert (changes requested):
Looks great overall, but just a (hopefully) small change to request.

PR #183, opened 2025-07-31
BULK scripts to help with infilling data in a consistent manner

Nospamas:
Adds scripts that allow looped processing over a time range, wrapping our existing download, process and pipeline utilities.

Scripts also in the directory may now be redundant and ripe for clean up. Specifically, these scripts should work with any network by passing through command arguments to the subroutines, network specific infilling shouldn't be necessary any more.


rod-glover (changes requested):

scripts/bulk_pipeline.py
```diff
@@ -0,0 +1,237 @@
...
+        log.info(
+            f"Time range processing complete. Success: {successful_runs}, Failed: {failed_runs}"
+        )
+
+    except Exception as e:
+        log.error(f"Error in time range processing: {str(e)}")
+        raise
+
+
+def generate_cache_filename(cache_dir, network, timestamp, frequency=None, tag=None):
```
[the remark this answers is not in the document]
rod-glover: See https://github.com/pacificclimate/crmprtd/blob/04eb0f70a5cf76f3fe329fd915eeecf4686b6ce9/crmprtd/download_cache_process.py#L86

crmprtd/download_cache_process.py
```diff
@@ -378,6 +396,7 @@ def main(arglist: List[str] = None) -> None:
 
         add_version_arg(parser)
 
+        # TODO: Overloaded argument, -D is for try run in other crmprtd utilities
         parser.add_argument(
             "-D",
             "--describe",
```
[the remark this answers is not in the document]
Nospamas: I think I'll leave both to a separate PR as it will end up changing the existing API, though I doubt the describe param will break any existing usage I'm still inclined to be cautious.

rod-glover (approved):
Looking good, a few minor comments.

crmprtd/infill.py
```diff
@@ -54,9 +54,9 @@ def download_and_process(
     network_name: str,
     log_args: List[str],
     download_args: List[str],
-    cache_filename: str = None,
-    connection_string: str = None,
-    dry_run: bool = None,
+    cache_filename: Optional[str] = None,
+    connection_string: Optional[str] = None,
+    dry_run: bool = True,
```
rod-glover: How about default `False`? There's no "not dry-run" except for `--dry_run=False`.

scripts/bulk_pipeline.py
```diff
@@ -11,13 +12,77 @@
...
+
+            # Replace the default cache directory with the user-specified directory
+            cache_filename = cache_filename.replace(
+                f"~/{opts.network}/cache/", f"{opts.directory}/{opts.network}/cache/"
+            )
+
+            # ensure directory exists
+            cache_dir = os.path.dirname(cache_filename)
+            if not os.path.exists(cache_dir):
+                os.makedirs(cache_dir, exist_ok=True)
```
rod-glover, line 44: This is much the same as `ensure_log_directory` above. DRY up?

scripts/bulk_pipeline.py
```diff
@@ -72,61 +136,26 @@ def main(opts, args):
...
+                    raise ValueError("Province must be specified for EC network")
+                # Process each province if specified
+                if opts.network == "ec":
+                    for p in opts.province:
+                        log.info(f"Processing province: {p}")
+                        copts = copy.copy(opts)
+                        copts.province = p.lower()  # Normalize to lowercase
+                        process(current_time, copts, args)
+                        successful_runs += 1
+                        log.info(f"Successfully processed: {iter_time_str} for province {p}")
```
rod-glover, line 152: :+1:

scripts/bulk_pipeline.py
```diff
@@ -205,9 +229,14 @@ def main(opts, args):
 
     opts, args = parser.parse_known_args(sysargs)
 
+    if opts.network == "ec":
+        add_province_args(parser)
+        opts, args = parser.parse_known_args(sysargs)
+        # Normalize to lowercase.
+        opts.province = [p.lower() for p in opts.province]
```
rod-glover: :+1:

rod-glover (approved):
LGTM, except for default val for dry_run

PR #185, opened 2025-08-25
update package so that it includes new scripts, entrypoints

Nospamas:
Quick PR to update and include bulk scripts in packages

### DIFF
--- a/crmprtd/__init__.py
+++ b/crmprtd/__init__.py
@@ -236,6 +236,7 @@ def setup_logging(log_conf, log_filename, error_email, log_level, name):
 def subset_dict(a_dict, keys_wanted):
     return {key: a_dict[key] for key in keys_wanted if key in a_dict}
 
+
 def add_province_args(parser):
     # See notes below re. province codes.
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -32,6 +32,8 @@ crmprtd_pipeline = "crmprtd.download_cache_process:main"
 crmprtd_download = "crmprtd.download:main"
 crmprtd_process = "crmprtd.process:main"
 crmprtd_gulpy = "crmprtd.process:gulpy_plus_plus"
+bulk_pipeline = "scripts.bulk_pipeline:run"
+bulk_process = "scripts.bulk_process:run"
```
rod-glover, line 36: How do you feel about continuing the "standard" naming of `main` for this method, not `run`?

rod-glover (approved):
LGTM, with one question and assuming that you have tried it out and it runs.

pyproject.toml
Nospamas: done :)

